PR #18, opened 2020-03-06
Error handling and await to added

TeddyAndersson:


services/bankid-api/lambdas/collect.js
```diff
@@ -20,7 +20,7 @@ export const main = async event => {
       bankId.url(bankidSSMParams.apiUrl, '/collect'),
       payload,
     );
-
+    console.log(data);
```
dannil76: Do we need this?

PR #29, opened 2020-03-17
Feature/userdata api

dannil76:
add and get user (mock)


dannil76:
> Service throws bad gateway error on both get and post requests. Locate issues and fix the services

Done!

PR #33, opened 2020-03-22
Fix/cleanups bankid api

dannil76:
Cleaned up technical debt


TeddyAndersson:
The certificates from S3 cannot be collected due to changes in the ssm json envs. Fix example template for bankidEnvs in resource repo.

TeddyAndersson:
Works fine now (y)

PR #34, opened 2020-03-22
Fix/cleanups user api

dannil76:
Cleaned up technical debt


services/user-api/lambdas/get.js
```diff
@@ -10,7 +11,7 @@ export const main = async event => {
   const { personalNumber } = event.pathParameters;
 
   const params = {
-    TableName: 'users',
+    TableName: config.users.TableName,
```
TeddyAndersson: Are AWS aware of this config params? How does these config vars get compiled with serverless?

TeddyAndersson: Asking this since i could not retrive a user from my database after sls deploy.

dannil76: Fixed typo. Should be config.users.tableName.

PR #35, opened 2020-03-22
Feature/mock data navet ms

dannil76:


services/navet-ms/mock/data.json
```diff
@@ -0,0 +1,5 @@
+{
+  "detail": {
+    "personalNumber": "195809262743"
```
TeddyAndersson: personalNumber should be an int/number?

dannil76: personalNumber changed to integer data type

PR #44, opened 2020-04-01
Added more precise error handling to all bankId requests

TeddyAndersson:


libs/response.js
```diff
@@ -2,16 +2,20 @@ export function success(body) {
   return buildResponse(200, body);
 }
 
-export function failure(error) {
+export function failure(error, stackTrace) {
+  const { stage } = process.env;
```
dannil76, line 6: we may need to use config.js to be consistent?

services/bankid-api/serverless.yml
```diff
@@ -72,5 +72,5 @@ functions:
     events:
       - http:
           path: auth/bankid/cancel
-          method: delete
+          method: post
```
dannil76, line 75: why not delete?

libs/response.js
TeddyAndersson: Agree, forgot about config. Good Catch 🙂

services/bankid-api/serverless.yml
TeddyAndersson: Becouse the BankID docs uses post, and since we reference ro that documentation.

Would also argue that we do not remove anything, it's an abort/cancel operation 🙂

dannil76: good point

PR #46, opened 2020-04-16
BankId Sign

mysticlagoona:
updated code
 added error msg in auth

### DIFF
--- a/services/bankid-api/lambdas/auth.js
+++ b/services/bankid-api/lambdas/auth.js
@@ -20,6 +20,7 @@ export const main = async event => {
   };
 
   const [error, bankIdAuthResponse] = await to(sendBankIdAuthRequest(bankIdSSMparams, payload));
+
   if (!bankIdAuthResponse) return response.failure(error);
 
   return response.success({
@@ -34,7 +35,8 @@ async function sendBankIdAuthRequest(params, payload) {
   let error, bankIdClientResponse, bankIdAuthResponse;
 
   [error, bankIdClientResponse] = await to(bankId.client(params));
-  if (!bankIdClientResponse) throwError(503);
+
+  if (!bankIdClientResponse) throwError(503, error);
 
   [error, bankIdAuthResponse] = await to(
     request.call(bankIdClientResponse, 'post', bankId.url(params.apiUrl, '/auth'), payload)
--- a/services/bankid-api/lambdas/auth.js
+++ b/services/bankid-api/lambdas/auth.js
@@ -20,7 +20,6 @@ export const main = async event => {
   };
 
   const [error, bankIdAuthResponse] = await to(sendBankIdAuthRequest(bankIdSSMparams, payload));
-
   if (!bankIdAuthResponse) return response.failure(error);
 
   return response.success({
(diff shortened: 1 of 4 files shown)


services/bankid-api/lambdas/auth.js
```diff
@@ -1,7 +1,6 @@
 import to from 'await-to-js';
```
TeddyAndersson, line 1: How come you made changes to this function?

services/bankid-api/lambdas/sign.js
```diff
@@ -1,42 +1,59 @@
 import { failure, success } from '../../../libs/response';
 import params from '../../../libs/params';
+import config from '../../../config';
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
 import * as request from '../../../libs/request';
 import * as bankId from '../helpers/bankId';
+import to from 'await-to-js';
+import snakeCaseKeys from 'snakecase-keys';
```
dannil76, line 8: We should use camel case.

services/bankid-api/lambdas/sign.js
```diff
@@ -1,42 +1,59 @@
...
-    return failure({ status: false, error: error.message });
+  if (!endUserIp || !userVisibleData) {
+    const error = {
+      status: 400,
+      code: 400,
+      name: 'BadRequestError',
+      detail: 'The server cannot process the request',
+      message: 'Missing endUserIp or userVisibleData',
+    };
+    return failure(error);
```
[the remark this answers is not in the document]
TeddyAndersson: would be nice if we could make a more general validation utility function but that is out of scope for this PR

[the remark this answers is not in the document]
TeddyAndersson: Or maybe not? 
Could you try to write a function that takes a callback function where we pass the throwError function?

My thought is that it could be used like this:
` async validateEventBody(error, callbackFunction(event.body)) {
   ...
   if(!callbackFunction()){
       throwError(...error)
   }
  ....
}`

services/bankid-api/lambdas/auth.js
mysticlagoona: Niko found it as was checking the error msgs when I was trying to  connect.

services/bankid-api/lambdas/sign.js
```diff
@@ -1,42 +1,68 @@
...
+
+  [error, bankIdSignResponse] = await to(
+    request.call(bankIdClientResponse, 'post', bankId.url(params.apiUrl, '/sign'), payload)
+  );
+
+  if (!bankIdSignResponse) throwError(error.response.status, error.response.data.details);
+  return bankIdSignResponse;
+}
+
+export const main = async event => {
```
TeddyAndersson, line 42: A minor thing. But we have the export of Main functions at the top in all our other services. So to be consistent i think se should move this to the top aswell. 

However i don't know if this is better acctually. Feels more natural to export at the end of the file 😄

mysticlagoona: Normally I keep the export function in the end and sub function on top. :)

But I changed it to top for consistency.

TeddyAndersson: Yeah maybe we should discuss the export syntax (y)

PR #48, opened 2020-04-21
JWT based auhtentication and authorization flow

TeddyAndersson:
A first implementation of a jwt based authentication/authorization flow. 

How it works:
The client that want to authenticate a user needs to have a valid apiKey, this is generated by simply running the serverless deploy command.

The apiKey can further be used to generate tokens from the auth/token endpoint. This is done by sending a POST request with any JSON Body and a x-api-key header with the apiKey as a value.

After this is done the client should now have a valid accessToken and can make requests to any endpoint that are restricted with a token authentication. (Currently there is none implemented)

To test this implementation you can create a dummy lambda in the auth/token directory that hooks on to the authorizer function.

### DIFF
--- a/services/auth/token/helpers/generateIAMPolicy.js
+++ b/services/auth/token/helpers/generateIAMPolicy.js
@@ -1,0 +1,21 @@
+// Helper function to generate an IAM policy
+const generateIAMPolicy = function(principalId, effect, resource) {
+  var authResponse = {};
+
+  authResponse.principalId = principalId;
+  if (effect && resource) {
+    var policyDocument = {};
+    policyDocument.Version = '2012-10-17';
+    policyDocument.Statement = [];
+    var statementOne = {};
+    statementOne.Action = 'execute-api:Invoke';
+    statementOne.Effect = effect;
+    statementOne.Resource = resource;
+    policyDocument.Statement[0] = statementOne;
+    authResponse.policyDocument = policyDocument;
+  }
+
+  return authResponse;
+};
+
+export default generateIAMPolicy;
--- a/services/auth/token/helpers/token.js
+++ b/services/auth/token/helpers/token.js
@@ -1,0 +1,18 @@
+import to from 'await-to-js';
+import jwt from 'jsonwebtoken';
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
+
+const SECRET_KEY = 'secretKeey';
+
+export const signToken = jsonToSign => {
+  const token = jwt.sign(jsonToSign, SECRET_KEY);
+  return token;
+};
+
+export const verifyToken = async token =>
+  jwt.verify(token, SECRET_KEY, (error, decoded) => {
+    if (error) {
+      throwError(401, error.message);
+    }
+    return decoded;
+  });
--- a/services/auth/token/lambdas/authorize.js
+++ b/services/auth/token/lambdas/authorize.js
@@ -1,0 +1,19 @@
+import to from 'await-to-js';
+
+import { verifyToken } from '../helpers/token';
+import generateIAMPolicy from '../helpers/generateIAMPolicy';
+
+export const main = async event => {
+  const { authorizationToken, methodArn } = event;
+  let isAllowed = true;
+
+  const [error] = await to(verifyToken(authorizationToken));
+  if (error) {
+    isAllowed = false;
+  }
+
+  const effect = isAllowed ? 'Allow' : 'Deny';
+  const IAMPolicy = generateIAMPolicy('user', effect, methodArn);
+
+  return IAMPolicy;
+};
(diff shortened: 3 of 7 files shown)


services/auth/token/helpers/generateIAMPolicy.js
```diff
@@ -0,0 +1,21 @@
+// Helper function to generate an IAM policy
+const generateIAMPolicy = function(principalId, effect, resource) {
```
dannil76, line 2: function(...
function (...
According to my linter we're missing space

services/auth/token/helpers/generateIAMPolicy.js
```diff
@@ -0,0 +1,21 @@
...
+  var authResponse = {};
```
dannil76, line 3: why usage var ?!

services/auth/token/helpers/generateIAMPolicy.js
```diff
@@ -0,0 +1,21 @@
...
+
+  authResponse.principalId = principalId;
+  if (effect && resource) {
+    var policyDocument = {};
```
dannil76, line 7: var?!

services/auth/token/helpers/generateIAMPolicy.js
```diff
@@ -0,0 +1,21 @@
...
+    policyDocument.Version = '2012-10-17';
+    policyDocument.Statement = [];
+    var statementOne = {};
```
dannil76, line 10: var is bad

services/auth/token/helpers/token.js
```diff
@@ -0,0 +1,18 @@
+import to from 'await-to-js';
```
dannil76, line 1: never used in this file?

services/auth/token/helpers/token.js
```diff
@@ -0,0 +1,18 @@
...
+import jwt from 'jsonwebtoken';
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
+
+const SECRET_KEY = 'secretKeey';
+
+export const signToken = jsonToSign => {
```
dannil76, line 7: signToken = jsonToSign => {...
signToken = (jsonToSign) => {...
according to my linting rules

services/auth/token/helpers/token.js
```diff
@@ -0,0 +1,18 @@
...
+  const token = jwt.sign(jsonToSign, SECRET_KEY);
+  return token;
+};
+
+export const verifyToken = async token =>
```
dannil76, line 12: verifyToken = async token =>...
verifyToken = async (token) =>...
 according to my linting rules

services/auth/token/lambdas/authorize.js
```diff
@@ -0,0 +1,19 @@
+import to from 'await-to-js';
+
+import { verifyToken } from '../helpers/token';
+import generateIAMPolicy from '../helpers/generateIAMPolicy';
+
+export const main = async event => {
```
dannil76, line 6: main = async event => {...
main = async (event) => {...
according to my linting rules

services/auth/token/lambdas/generateToken.js
```diff
@@ -0,0 +1,13 @@
+import { signToken } from '../helpers/token';
+
+export const main = async event => {
```
dannil76, line 3: main = async event => {...
main = async (event) => {...
according to my linting rules

services/auth/token/helpers/generateIAMPolicy.js
TeddyAndersson: This is bad, really bad

services/auth/token/helpers/token.js
TeddyAndersson: What, i would say that no parameters is the way if there is only one argument passed down.

services/auth/token/helpers/generateIAMPolicy.js
TeddyAndersson: This should be an arrow function, for the sake of it

services/auth/token/helpers/generateIAMPolicy.js
```diff
@@ -0,0 +1,21 @@
...
+    statementOne.Effect = effect;
+    statementOne.Resource = resource;
+    policyDocument.Statement[0] = statementOne;
+    authResponse.policyDocument = policyDocument;
+  }
+
+  return authResponse;
+}
+
+export default generateIAMPolicy;
```
dannil76, line 21: you are exporting the function name in file but not in other files? Should aim for consistently.

TeddyAndersson: I hear you and the changes has been made. Typical component exports is done like this, so probably did this from react experience :P

PR #49, opened 2020-04-27
Feature/token request validation

TeddyAndersson:
Validation of the incoming request to check if the key personalNumber exists and that it's value is a valid SSN (Swedish Social Security Number).

### DIFF
--- a/services/auth/token/lambdas/generateToken.js
+++ b/services/auth/token/lambdas/generateToken.js
@@ -1,13 +1,65 @@
+import to from 'await-to-js';
+
+import * as response from '../../../../libs/response';
 import { signToken } from '../helpers/token';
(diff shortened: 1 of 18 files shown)


services/auth/token/lambdas/generateToken.js
```diff
@@ -1,13 +1,65 @@
...
+
+  return [true];
+}
+
+/**
+ * Check if a string is a valid SSN(Swedish Social Security Number)
+ * @param {string} ssn
+ */
+function isSwedishSocialSecurityNumber(ssn) {
+  if (/^(19|20)?(\d{6}(-|\s)\d{4}|(?!19|20)\d{10})$/.test(ssn)) {
```
dannil76, line 52: Accept number with T inside?

PR #52, opened 2020-04-28
Secret for signing tokens is retrived from AWS secrets manager

TeddyAndersson:


services/auth/token/helpers/token.js
```diff
@@ -1,11 +1,12 @@
 import jwt from 'jsonwebtoken';
 import { throwError } from '@helsingborg-stad/npm-api-error-handling';
+import secrets from '../../../../libs/secrets';
 
-// TODO: Retrive and use real SECRET_KEY from AWS;
-const SECRET_KEY = 'secretKeey';
+const SECRET_KEY = secrets.get('dev/token/generation', 'TokenGeneratorSecret');
```
dannil76, line 5: Put in config.js

services/auth/token/lambdas/generateToken.js
```diff
@@ -2,8 +2,7 @@ import { signToken } from '../helpers/token';
 
 export async function main(event) {
   const jsonRequest = JSON.parse(event.body);
-  const token = signToken(jsonRequest);
-
+  const token = await signToken(jsonRequest);
```
dannil76, line 5: Await to?

services/auth/token/helpers/token.js
```diff
@@ -1,11 +1,12 @@
...
 
-export function signToken(jsonToSign) {
-  const token = jwt.sign(jsonToSign, SECRET_KEY);
+export async function signToken(jsonToSign) {
+  const secret = await SECRET_KEY;
```
dannil76, line 8: Await to?

libs/secrets.js
```diff
@@ -0,0 +1,31 @@
+import AWS from 'aws-sdk';
```
dannil76, line 1: Should import our custom aws-sdk pkg to support future features

TeddyAndersson: Don't se why we should use the aws-sdk pkg defined in /libs, it currently does nothing more than returning the aws import? Or for that manner why, we have added a wrapper somewhere along the line?

TeddyAndersson: Just don't get the reason for having a wrapper, and i can see know that we have been using it mixed with normal aws-sdk imports. 

So some files are importing from 'aws-sdk' and others from './aws-sdk'

dannil76: remove the wrapper for now then

PR #70, opened 2020-07-08
Feat/update form

zilaei:
Added logic for updating form, step & question.

### DIFF
--- a/services/forms-api/lambdas/updateQuestion.js
+++ b/services/forms-api/lambdas/updateQuestion.js
@@ -2,7 +2,7 @@ import to from 'await-to-js';
 
 import config from '../../../config';
 import * as response from '../../../libs/response';
-import { createUpdateExpression, updateItem } from '../helpers/queries';
+import { updateItem } from '../helpers/queries';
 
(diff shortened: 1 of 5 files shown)


services/forms-api/lambdas/updateQuestion.js
```diff
@@ -0,0 +1,38 @@
...
+import { updateItem } from '../helpers/queries';
+
+export async function main(event) {
+  const { formId, stepId, questionId } = event.pathParameters;
+  const questionPartitionKey = `FORM#${formId}`;
+  const questionSortKey = `${questionPartitionKey}#STEP#${stepId}#QUESTION#${questionId}`;
+
+  const requestBody = JSON.parse(event.body);
+
+  const validKeys = ['label', 'description', 'show', 'type'];
```
jwinding, line 14: For the questions, we want to allow any key, because different types of inputs requires different additional props. So here any kind of key (possibly excluding some, like PK and SK) should be accepted and updated.

jwinding:
Closing, since we seem to be sticking with the other form-api.

PR #71, opened 2020-07-13
Feat/forms3

jwinding:
Adds a new and simplified forms api, that makes it much easier to create and update forms.


jwinding:
approved by Ehsan in discussion.

PR #75, opened 2020-07-24
Store current page number in db

jwinding:
Small change to the update function so that it saves the currentPage if that is sent in as part of an update of a case.

### DIFF
--- a/services/cases-api/lambdas/update.js
+++ b/services/cases-api/lambdas/update.js
@@ -26,16 +26,16 @@ export async function main(event) {
     ExpressionAttributeNames['#status'] = 'status';
     ExpressionAttributeValues[':newStatus'] = requestBody.status;
   }
-  if (requestBody.status && requestBody.currentPage) UpdateExpression += ', ';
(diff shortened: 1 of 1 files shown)


jwinding:
Now the logic should work as intended, and it also returns the proper thing.

services/cases-api/lambdas/update.js
```diff
@@ -26,10 +26,16 @@ export async function main(event) {
     ExpressionAttributeNames['#status'] = 'status';
     ExpressionAttributeValues[':newStatus'] = requestBody.status;
   }
+  if (requestBody.status && requestBody.currentPage) UpdateExpression += ', ';
+  if (requestBody.currentPage) {
+    UpdateExpression += '#currentPage = :newPage';
+    ExpressionAttributeNames['#currentPage'] = 'currentPage';
+    ExpressionAttributeValues[':newPage'] = requestBody.currentPage;
+  }
```
TeddyAndersson, line 34: currentPage should be renamed to currentStep

PR #76, opened 2020-08-11
fix(user-api): add 404 message, and format response to camelcase

jwinding:
Fixes the user-api so that it returns 404 and an error message if the user is not found in the db, and formats the response (if the user was found) to be in camelCase.

Also include an expiration time in the JWT token that is generated at auth/token.

### DIFF
--- a/services/auth/token/lambdas/generateToken.js
+++ b/services/auth/token/lambdas/generateToken.js
@@ -11,6 +11,9 @@ export const main = async event => {
   );
   if (error) return response.failure(error);
 
+  // Add expiration time to JWT token, 15 min.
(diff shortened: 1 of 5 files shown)


services/auth/token/lambdas/generateToken.js
```diff
@@ -11,6 +11,9 @@ export const main = async event => {
   );
   if (error) return response.failure(error);
 
+  // Add expiration time to JWT token, 15 min.
+  // The format is in seconds since Jan 1, 1970, not milliseconds, to match the default iat format of JWT.
+  validatedEventBody.exp = parseInt(Date.now() / 1000) + 15 * 60;
```
TeddyAndersson, line 16: Move to signToken function

PR #77, opened 2020-08-13
Fix/import authorizer

TeddyAndersson:
This PR fixes the issue of not being able to import the Custom Authorizer Lambda on more than one service. The problem in itself was due to that in order to reference the deployed authorizer it must be imported through the id of the lambda. If the import of an authorizer is made from the authorizers arn the import will try to deploy the custom authorizer and attach it to the ApiGateway that already has a an authorizer connected to it.

The Solution to this problem was to export the Authorizer id and tell the ApiGateway that we want to use this specific Lambda to handle Authorization. Then when we want to import the Authorizer in a service we can do a direct reference in its ID, which will prevent the service to setup it's own Authorizer in the stack once we deploy it.

### DIFF
--- a/services/auth/token/serverless.yml
+++ b/services/auth/token/serverless.yml
@@ -41,7 +41,7 @@ provider:
       Resource: "*"
 
 functions:
-  authorize:
+  authorizer:
     handler: lambdas/authorize.main
 
   generateToken:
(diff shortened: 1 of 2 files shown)


TeddyAndersson:
> Looks good, and I think I understand the overall logic, even though it took a while.

Yeah it's not that intuitive :P
Did you test deploy it in your aws enviorment? :)

TeddyAndersson:
Could someone test deploy the auth/token, case-api and user-api service in order for me to know that this solution works on antoher aws instance than just mine.

What you need to test is that you can generate a token and use it to send a get request on the user and the cases endpoint

jwinding:
Har deployat och testat på min AWS miljö, vad jag kan se funkar det som det ska.

PR #79, opened 2020-08-14
fix(update): update case now adds an updatedAt

jwinding:
update case now adds an updatedAt, and returns it.


TeddyAndersson:
Looks good to me (y)

PR #80, opened 2020-08-20
Fix/cases get pno from token

TeddyAndersson:
Before the Authorization header was used completely incorrect. The Authorization header was used to pass down a userId that later on was used to get cases from dynamodb. 

Now the Authorization header is used to pass down a valid accessToken from then client. Then we decode the token to get the payload from the token which determine the cases that we should get from AWS DynamoDB. With other words you can't get a user without a valid accessToken with this update.

### DIFF
--- a/services/cases-api/lambdas/list.js
+++ b/services/cases-api/lambdas/list.js
@@ -1,4 +1,5 @@
 import to from 'await-to-js';
+import jwt from 'jsonwebtoken';
 import { throwError } from '@helsingborg-stad/npm-api-error-handling';
 
 import config from '../../../config';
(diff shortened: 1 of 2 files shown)


jwinding (changes requested):
It works when using a valid token, but when using an invalid or expired token, you get a { message: null } and a status 500 internal server error. This should be changed to a more descriptive error and a correct error code.

TeddyAndersson:
This is due to the authorizer, are looking into this right now. This PR is with other words not the cause of the authorization errors, that's the authorizer. So if this works with a valid token it's acting as intended. (y)

jwinding (approved):
This code works and can be merged, since the issue was in the authorizer as stated.

PR #81, opened 2020-08-20
fix: get user from token payload

TeddyAndersson:
Before when a get request was made towards users/{personalNumber} we just returned the user that was requested if the accessToken was valid. This is however a problem, since it means that anyone with a valid token can request any user, if they passed the correct personalNumber in the path.

To fix this issue we are now relaying on the payload of the jwt token to pass the personalNumber instead of retrieving it from the path. This means that the path is currently completely useless, but since the application Mitt Helsingborg uses it we keep the personalNumber in the path. This will however be replaced in a near future with a new endpoint.

### DIFF
--- a/services/user-api/lambdas/get.js
+++ b/services/user-api/lambdas/get.js
@@ -1,4 +1,5 @@
 import to from 'await-to-js';
+import jwt from 'jsonwebtoken';
 import camelCase from 'camelcase';
 import { throwError } from '@helsingborg-stad/npm-api-error-handling';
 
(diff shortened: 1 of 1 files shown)


jwinding (changes requested):
Same thing here as in the other recent PR: it works when using a valid token, but when the token is expired or invalid you get a null response and a code 500 error. We should try and have at least somewhat helpful error messages.

TeddyAndersson:
This is due to the authorizer, are looking into this right now. This PR is with other words not the cause of the authorization errors, that's the authorizer. So if this works with a valid token it's acting as intended. (y)

jwinding (approved):
Same thing, this works, the problem is in the authorizer. So make a new pull request for fixing that, this is fine.

PR #82, opened 2020-08-20
feat: new endpoint me on users-api

TeddyAndersson:
New endpoint users/me will replace users/{personalNumber}.


jwinding (approved):
Good. I like the naming of /users/me, that seems very clear.

PR #83, opened 2020-08-20
fix(cases): flatten response objects from cases-api

jwinding:
In the cases api, we were returning objects with the structure {type, id, attributes:{ data:{}, ...}}. Then in the app, we flatten these objects out, removing the attributes. So this is pretty stupid and as far I can see, separating some things into attributes serves no purpose, so this changes it to be flat from the beginning.

### DIFF
--- a/services/cases-api/lambdas/create.js
+++ b/services/cases-api/lambdas/create.js
@@ -48,14 +48,14 @@ export async function main(event) {
   if (dynamodbError) return response.failure(dynamodbError);
 
   return response.success(201, {
-    type: 'cases',
     id: caseId,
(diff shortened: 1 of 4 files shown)


TeddyAndersson:
You might se this as stupid. But the team have agree upon following the jsonapi standard in our api responses. You can find the standard for jsonapi [here](jsonapi.org]). If we would agree to change this response we also need to change all other responses in the api, to be consistent. So for now i will not approve this PR, however i'm open for a discussion of changing standard format for the response.

TeddyAndersson:
Closing this until a discussion regarding standards have been held.

PR #85, opened 2020-08-20
fix: authorizer ruturns a 401 when unauthorized

TeddyAndersson:
Earlier when a token was invalid the error response on a request was 500 internal server error.
This issue is now solved so that if the token is invalid a 401 unauthorized response is returned.


jwinding (approved):
Good, this works.

PR #86, opened 2020-08-24
Get pnr from token in cases-api

jwinding:
Updates the cases-api to get the Personal number from the decoded token rather than reading it from plaintext in the header or body. 
Note that when deploying this to the dev-AWS, the app will need to be changed in order to work, but the changes should be only to the case context. 

Also refactored out the decoding of the token into a separate function placed in libs, since it's used across multiple services.

### DIFF
--- a/libs/token.js
+++ b/libs/token.js
@@ -1,0 +1,16 @@
+import jwt from 'jsonwebtoken';
+
+/**
+ * Takes an http event with an jwt authorization header, and returns the decoded info from it. Does not check if the token is valid, that should be handled by an authorizer.
+ * @param {*} httpEvent the event passed to the lambda
+ */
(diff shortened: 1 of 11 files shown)


TeddyAndersson:
We also need to add the authorizer in serverless.yml on all endpoints for this to work

TeddyAndersson (changes requested):
Add authorizer on http event in serverless.yml

jwinding:
> We also need to add the authorizer in serverless.yml on all endpoints for this to work

Well, no, it works anyways I think; you can send a token and decode it even if it's not behind an authorizer. However all the endpoints should probably be behind the authorizer, so it's good to do this anyways.

TeddyAndersson (approved):
Looks good (y)

PR #87, opened 2020-08-24
feat(forms3): add api key to forms3 and makes it private

jwinding:
Hides all the endpoints of the forms3 api behind a new apikey, specific to this api. This is done in part because obviously this api should not be fully public, and in part to have an easy method of securing the form builder. 

Note that on deploying this, we need to change the app so that it sends this key when getting forms. 
Later we might want to use the same api key, but this is a little harder to setup, so for now lets use this.


jwinding:
Changed it so that it uses the same api key, using a serverless plugin made for this purpose. This is specified in the serverless.common.yml, so to enable this key on a service one just turns on the plugin for that service, and set the functions that one wants to hide to private.

TeddyAndersson (approved):
Since we already have touched on this topic through teams, i will approve this. Good work (y)

PR #89, opened 2020-08-25
fix: checking if authorization header contains bearer prefix or not

TeddyAndersson:
When we send tokens in the Authorization header we would like to prefix them with the token type, in our case Bearer. This was not taken into consideration when implementation the custom authorizer lambda and therefore an Authorization header with Bearer as prefix would be treated as an invalid token.

This issue is now fixed and the custom authorizer does allow the Authorization header to be prefixed with Bearer. The prefix can also be excluded.


jwinding (approved):
Looks good.

PR #90, opened 2020-08-26
refactor: moving token helper functions to libs

TeddyAndersson:
Helper functions for token verification and signature have been move to libs.


jwinding (approved):
Good.

PR #92, opened 2020-09-07
Feat/s3 signed upload url

TeddyAndersson:
Before there was no way to upload images to AWS S3 through the api gateway.

This PR makes it possible to request a pre signed upload url that can be used in a client to do direct uploads towards a bucket in AWS S3.

This PR does support any type of file sizes to be uploaded, which will need to be solved in a future implementation.

NOTE: To test this PR you need to setup the attachments bucket that are found in the resources repo, otherwise the added attachments endpoint will not work.

Test this PR

** 1. Send a POST request towards the endpoint users/me/attachments with the following body:**
`{ 
    "fileName": "somefilename", 
    "mime": "image/jpg" // (could also be /png or /jpeg)
}`
This will grant you a response with a pre signed upload url.

** 2. Upload a file with the url**
Send a PUT request towards the pre signed URL, in the request body you will need to attach a binary file that matches the mime that you earlier entered to obtain the pre signed url

** 3. Check for file in S3**
The last step you will need to do is to check if the file have been uploaded, this can be done through the aws management console at aws.amazon.com-

### DIFF
--- a/libs/S3.js
+++ b/libs/S3.js
@@ -16,46 +16,6 @@ async function getSignedUrl(bucketName, method, params) {
   });
 }
 
-/**
- * Use AWS SDK to create pre-signed POST data.
- * We also put a default file size limit (100B - 10MB).
- * @param bucketName
- * @param key
- * @param contentType
- * @param expireTime
- * @param fileMinSize
- * @param fileMaxSize
- * @returns {Promise<object>}
- */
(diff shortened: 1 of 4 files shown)


jwinding (approved):
Working as intended. Tested it from my app component, and the file gets correctly uploaded.

PR #98, opened 2020-09-17
fix: moved retrival of secrets for token inside of functions

TeddyAndersson:
Due to policy/access problem. This issue was causing the api to act weird when creating and updating a case. This was due to a retrieval of secret keys from aws secrets manager in the background without the correct aws policy credentials


jwinding (approved):
Good!

dannil76:
Take the freedome to rebase into dev

PR #102, opened 2020-10-12
fix: remove requirement of step description, add step id as required

jwinding:
## Feature description
Changes what fields are required on steps in a form: the step description should be allowed to be empty, and with the new stepper logic, each step needs to have its own id (generated uniquely in the form builder). 

## Solution description
Changes the formValidation to incorporate these changes. 

The formValidation helper file.

## Is there any existing behaviour change of other features due to this code change?
Only what is required when creating and updating a form. 

No.

## Are your code structured in a way so that reviewers can understand it?
Yes.

- [X] Your personal AWS environment.


dannil76 (approved):
Works as intended

PR #103, opened 2020-10-15
Fix/cleanup

jonatanhanson:
- Renamed services, methods etc. for better consistency
- Fixed lint warnings
- Fixed some Readmes
- Moved helpers to libs
- Removed Snakecase lib
- Removed unused services (Watson)

Needs to be tested together with this PR:
https://github.com/helsingborg-stad/helsingborg-io-sls-resources/pull/33

This will break a lot of things so you need to remove every resources and services from your AWS and deploy again. 😱


helsingborgsstad:
Task linked: [CU-90vyx1 Städning AWS Api och Resources](https://app.clickup.com/t/90vyx1)

dannil76 (changes requested):
Got error when deploying with new stage name: services/auth/token
_An error occurred: ApiGatewayApiKey1 - helsingborg-io-apikey already exists in stack..._

Maybe add stage param to serverless.yml at line 19:
`- ${self:custom.stage}-helsingborg-io-apikey`

TeddyAndersson:
> Got error when deploying with new stage name: services/auth/token
> _An error occurred: ApiGatewayApiKey1 - helsingborg-io-apikey already exists in stack..._
> 
> Maybe add stage param to serverless.yml at line 19:
> `- ${self:custom.stage}-helsingborg-io-apikey`

Got the same issue

TeddyAndersson:
> > Got error when deploying with new stage name: services/auth/token
> > _An error occurred: ApiGatewayApiKey1 - helsingborg-io-apikey already exists in stack..._
> > Maybe add stage param to serverless.yml at line 19:
> > `- ${self:custom.stage}-helsingborg-io-apikey`
> 
> Got the same issue

When deploying the forms service

TeddyAndersson:
Everything deployed well except from the apiKey on Forms. Maybe we should take a look at how the key is generated, feels like it should be generated from the resources and not on a single service :)

PR #106, opened 2020-10-28
feat: 🎸 refactor create case lambda to reflect data struct

dannil76:
BREAKING CHANGE: 🧨 YES
Old form data structure will hurt you badly


jonatanhanson (approved):
Works for me!

PR #108, opened 2020-11-10
fix: 🐛 replace param: applicant for personalNumber

dannil76:
to be consistent with the data flow from app to viva


TeddyAndersson (approved):
Looks fine to me 🙌

PR #118, opened 2020-11-25
build(commitizen): making the project commitizen friendly

TeddyAndersson:
The project had missing packages for usage of commitizen. The packages have been updated so that the
project now is commitizen friendly


dannil76 (approved):
tested by deleting helper.js
what i can see is my workflow unchanged so it's fine for me

PR #120, opened 2020-11-26
Feat/validation case data

TeddyAndersson:

Good example:
I've added new validation logic for POST requests is sent to the cases endpoint.

These changes makes request validation easier to handle compared to the old implementation.

## Explain your solution

This includes adding a validation library (Joi) where developers can declare validation schemas instead of dealing with if statements.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Deploy the case-api to your personal AWS environment.
3. Send the POST requests listed further down (create a new case) to the cases endpoint.
4. Read the errors for the POST requests and make sure they are understandable.

Malformed values.
```JSON 
{
    "formId": "12345" // not a valid guid/uuid,
    "provider": "VIV" // not a valid provider
}
```

Missing required object key
```JSON 
{
    "formId": "98657f95-f7e4-4fd4-84e0-e1d798aa2981"
}
```

None allowed key
```JSON 
{
    "formId": "98657f95-f7e4-4fd4-84e0-e1d798aa2981",
    "provider": "VIVA",
    "details": ""
}
```

Malformed JSON format
```JSON 
{
    "formId": "98657f95-f7e4-4fd4-84e0-e1d798aa2981",
    "provider": "VIVA",
}
```

Successful request
```JSON 
{
    "formId": "98657f95-f7e4-4fd4-84e0-e1d798aa2981",
    "provider": "VIVA"
}
```

- [x] Personal AWS Dev environment.

### DIFF
--- a/services/cases-api/lambdas/createCase.js
+++ b/services/cases-api/lambdas/createCase.js
@@ -29,14 +29,11 @@ export async function main(event) {
 
   // TODO: check if the passed formId exsists in the form dynamo table.
 
-  const TableName = config.cases.tableName;
-
   const { personalNumber } = decodedToken;
-  const { provider, formId } = validatedEventBody;
   const id = uuid.v4();
 
   const caseItem = {
-    TableName,
+    TableName: config.cases.tableName,
     Item: {
       PK: `USER#${personalNumber}`,
       SK: `USER#${personalNumber}#CASE#${id}`,
@@ -46,8 +43,7 @@ export async function main(event) {
       status: CASE_STATUS_ONGOING,
       details: {},
       answers: [],
-      provider,
-      formId,
+      ...validatedEventBody,
     },
   };
 
--- a/services/cases-api/lambdas/createCase.js
+++ b/services/cases-api/lambdas/createCase.js
@@ -5,131 +5,83 @@ import uuid from 'uuid';
 import config from '../../../config';
 
 import * as response from '../../../libs/response';
-import { validateEventBody } from '../../../libs/validateEventBody';
-import { validateKeys } from '../../../libs/validateKeys';
+
 import { decodeToken } from '../../../libs/token';
 import { putItem } from '../../../libs/queries';
+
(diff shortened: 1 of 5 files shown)


services/cases-api/lambdas/createCase.js
```diff
@@ -5,131 +5,83 @@ import uuid from 'uuid';
...
-  const updatedAt = Date.now();
+  const TableName = config.cases.tableName;
 
   const { personalNumber } = decodedToken;
-  const { provider, formId, status, details, answers } = validatedEventBody;
-
-  const PK = `USER#${personalNumber}`;
-  const SK = `USER#${personalNumber}#CASE#${id}`;
+  const { provider, formId } = validatedEventBody;
+  const id = uuid.v4();
```
dannil76, line 36: could be put into the caseItem obj to be consistent

services/cases-api/lambdas/createCase.js
```diff
@@ -5,131 +5,83 @@ import uuid from 'uuid';
...
+    return response.failure(validationError);
   }
 
-  const TableName = config.cases.tableName;
+  // TODO: check if the passed formId exsists in the form dynamo table.
 
-  const id = uuid.v1();
-  const createdAt = Date.now();
-  const updatedAt = Date.now();
+  const TableName = config.cases.tableName;
```
dannil76, line 32: could be put into the caseItem obj to be consistent

dannil76 (changes requested):

services/cases-api/lambdas/createCase.js
TeddyAndersson: If you refer to the id, it's used in the SK. So therefore it's defined outside of the caseItem :)

dannil76 (approved):

jwinding (approved):
Works as intended. Some more validation can/should probably be done against the answers object, which can also be sent in when creating a case, and that ties into having the same kind of validation on the updateCase method, but this is a great start. And the library seems good.

PR #122, opened 2020-11-27
fix(case): rename current-step to current-position

jwinding:
A renaming of a variable to reflect a change in the data we store: currentStep used to be a number, now it's a larger position object.


dannil76 (approved):
Take this into account when implementing validation

PR #123, opened 2020-11-30
Fix/create case joi schema

TeddyAndersson:
The validation when doing a create case request have been updated to better match the current usage of the endpoint.

The first implementation of the validation on a create case request did not consider the current behavior that the APP Mitt Helsingborg is expecting from the endpoint. Changes have therefore been made to suite the current stage of the APP Mitt Helsingborg.

## Explain your solution
The Joi validation schema have been extended to allow the passing of the answers attribute. The answers attribute have also got validation for each answers, since we expect a certain data structure for an answer that is saved in a case.

In addition to this there have also been a change in how we retrieve and return values from a created case item. The reason for this change is that a putItem operation on DynamoDB will never return a newly created item. Therefore a second request to retrieve the created case item have been added.

## How to test the changes?

Concrete example:
1. Checkout this branch
2. Deploy the case-api to your personal AWS environment
3. Send a Post request to the case endpoint of your AWS API.
4. The response should tell you what is missing in the request body in order to do a successful request.

- [ ] Your local Serverless environment.

## Anyhting else? (optional)
The create case endpoint should be used to create a case for the first time, any updates to a case should be using a update case request. With this structure in mind there seems to be a missmatch in the Mitt Helsingborg app/client where the create case is used to both update and create a case, this was not the intention from the start. 

Also if there are some attributes that is missing and should be allowed in the creation of a case, please leave a comment with a explanation to why these attributes should be allowed on case creation :+1

### DIFF
--- a/services/cases-api/lambdas/createCase.js
+++ b/services/cases-api/lambdas/createCase.js
@@ -22,13 +22,13 @@ export async function main(event) {
   const [validationError, validatedEventBody] = await to(
     validateCreateCaseRequestBody(event.body, caseValidationSchema)
   );
-
   if (validationError) {
     return response.failure(validationError);
   }
 
   // TODO: check if the passed formId exsists in the form dynamo table.
 
+  const { formId, provider, ...rest } = validatedEventBody;
   const { personalNumber } = decodedToken;
   const id = uuid.v4();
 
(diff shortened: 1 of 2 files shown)


services/cases-api/lambdas/createCase.js
```diff
@@ -22,43 +22,61 @@ export async function main(event) {
...
+  if (getItemError) {
+    return response.failure(getItemError);
+  }
 
   return response.success(201, {
     type: 'createCases',
-    attributes,
+    attributes: {
+      ...caseItem,
+    },
```
dannil76, line 79: What do I get from caseItem?

dannil76 (approved):

services/cases-api/lambdas/createCase.js
TeddyAndersson: Updated the definition of Item and response for better readability

dannil76 (approved):

PR #125, opened 2020-12-02
feat: 🎸 add expirationTime

dannil76:
New attribute expirationTime used by DynamoDB(TTL) feature to delete
case on set date

### DIFF
--- a/services/cases-api/lambdas/createCase.js
+++ b/services/cases-api/lambdas/createCase.js
@@ -1,17 +1,16 @@
 /* eslint-disable no-console */
 import to from 'await-to-js';
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
 import uuid from 'uuid';
 
 import config from '../../../config';
-
 import * as response from '../../../libs/response';
-
 import { decodeToken } from '../../../libs/token';
 import { getItem, putItem } from '../../../libs/queries';
 
 import caseValidationSchema from '../helpers/schema';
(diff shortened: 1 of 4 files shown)


services/cases-api/lambdas/createCase.js
```diff
@@ -39,13 +39,15 @@ export async function main(event) {
   const PK = `USER#${personalNumber}`;
   const SK = `USER#${personalNumber}#CASE#${id}`;
   const timestamp = Date.now();
+  const expirationTime = generateExpirationTime(72);
```
TeddyAndersson: This input of number 72 should be moved to a common place, since we will use this in other services as discussed. If you prefer to move this now or later is up to you @dannil76

services/cases-api/lambdas/createCase.js
```diff
@@ -87,6 +89,20 @@ export async function main(event) {
   });
 }
 
+/**
+ * @param {number} futureHours hours in the future
+ * @returns {number} a future date in seconds
+ */
+function generateExpirationTime(futureHours) {
```
TeddyAndersson, line 96: A more sutible naming here would be calculateCaseExpirationTimeInSeconds. Since generateExpirationTime does not give me a hint on what the scope of the expiration time is applied to.

TeddyAndersson (commented):
Minor comments that could be addressed, but it works as expected

services/cases-api/lambdas/createCase.js
TeddyAndersson: also the argument here is kind of vague, why not only call it hours?

PR #128, opened 2020-12-07
Fix/delete case

dannil76:
Prevents the user from deleting case that one does not own


jwinding (approved):
Checked it on my personal AWS environment, and it works when I remove a case that belongs to the user in the token, and it doesn't work when trying to delete a case belonging to someone else. The error message ("The conditional request failed") could be a bit clearer, but that is secondary.

zilaei (approved):

PR #131, opened 2020-12-23
feat(pdf-ms): add a pdf generator ms

jwinding:
# PDF generator microservice

This microservice generates pdfs from case data, and saves it (as a buffer) on the case. 
Check the readme for some explanation of the data flow and template system. 

## How to deploy
Be sure to first deploy the pdfTemplates resource in the helsingborg-io-sls-resources repo, so that the required S3 bucket exists.
Then to deploy it, just run yarn in the service folder, and then deploy with serverless as normal, sls deploy. 

In the folder templateExample there is two template files that exemplifies how the template system works. These should be put into the S3 bucket named "pdf-case-templates-s3-{some-id}" for the generator to work. 

## How to test
To trigger the service, you need to update a case, setting its status to 'submitted'. So you have to do this, either through the app, or through a direct put-request against the case api. To view the generated pdf, you can go to the S3 bucket "pdf-case-templates-..." and it should be in the /cases/-folder. It's also saved on the case in the db, but as a buffer so you have to do some conversion to be able to view it from there. 

Then beware that it will only generate a pdf for a case once; unless you go into the db and switch the pdfGenerated value to false.


services/pdf-ms/helpers/caseDataConverter.ts
```diff
@@ -0,0 +1,44 @@
...
+ * Converts case answers array to object.
+ * @param {array} answerArray the flat array that we want to convert to a nested tree structure
+ */
+export const convertAnswerArrayToObject = (
+  answerArray: { field: { id: string; tags: string[] }; value: string }[]
+): Record<string, any> => {
+  const caseObject: Record<string, any> = {};
+  if (!Array.isArray(answerArray)) {
+    return caseObject;
+  }
```
dannil76, line 21: Plenty of noice words like answerArray, caseObject etc.
.....................................................^^^^^^..........^^^^^^^

services/pdf-ms/helpers/caseDataConverter.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  answerArray.forEach(answer => {
+    const path = answer.field.id.split('.');
```
dannil76, line 24: path to what?

services/pdf-ms/helpers/caseDataConverter.ts
```diff
@@ -0,0 +1,44 @@
...
+    path.reduce((prev, pathPart, i) => {
+      if (!prev) {
```
dannil76, line 26: why not prev? what is prev? what do i need it for?

If you meen previous of something then please enlight me on what this previous is and what it's used for

services/pdf-ms/helpers/caseDataConverter.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * Checks if string is a numeric value
+ * @param {string} str
+ */
+const isNumeric = (str: string): boolean => {
+  if (typeof str !== 'string') return false;
```
dannil76, line 6: if i don't pass in string it states that it's not a number? is string always numbers? feels odd...
if i pass in a true number it should return true right?

func name is lying?!

services/pdf-ms/helpers/caseDataConverter.ts
```diff
@@ -0,0 +1,44 @@
...
+        } else {
+          prev[pathPart] = {};
+        }
+      }
+
+      return prev[pathPart];
+    }, caseObject);
+  });
+
+  return caseObject;
```
dannil76, line 43: noice word alert

services/pdf-ms/helpers/caseDataConverter.ts
```diff
@@ -0,0 +1,44 @@
...
+          prev[pathPart] = answer.value;
+        } else if (isNumeric(path[i + 1])) {
+          prev[pathPart] = [];
+        } else {
+          prev[pathPart] = {};
+        }
+      }
+
+      return prev[pathPart];
+    }, caseObject);
```
dannil76, line 40: noice word alert

services/pdf-ms/helpers/objects.ts
```diff
@@ -0,0 +1,14 @@
+export function getPropertyFromDottedString(obj: Record<string, any>, dottedString: string): any {
```
dannil76, line 1: what does Property stand for? What property do i get?

services/pdf-ms/helpers/objects.ts
```diff
@@ -0,0 +1,14 @@
...
+  const strArr = dottedString.split('.');
```
dannil76, line 2: Please use pronounceable variable names - strArr?!
Array of what sort of strings?

services/pdf-ms/helpers/objects.ts
```diff
@@ -0,0 +1,14 @@
...
+  const [res] = strArr.reduce(
```
dannil76, line 3: what is res?

services/pdf-ms/helpers/objects.ts
```diff
@@ -0,0 +1,14 @@
...
+      if (!prev[0] && !prev[1]) return [undefined, false];
+      if (!prev[0] && prev[1] && obj[current]) return [obj[current], true];
+      if (prev[0] && prev[1] && prev[0][current]) return [prev[0][current], true];
+      if (prev[0] && prev[1] && !prev[0][current]) return [undefined, false];
+      return [undefined, true];
+    },
+    [undefined, true]
+  );
+  return res;
+}
```
dannil76, line 14: Sorry, to many WTF/minutes... :)

services/pdf-ms/helpers/pdf.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  const fonts = {
+    courier: courierFont,
+    timesRoman: timesRomanFont,
+    helvetica: helveticaFont,
+  };
+  return fonts;
+};
+
+/** Draws the text as specified in the textObject on top of the pdf page */
```
dannil76, line 18: maybe tell me this with a better func name then

services/pdf-ms/helpers/pdf.ts
```diff
@@ -0,0 +1,72 @@
...
+      return prev.replace(currentRegexResult[0], '');
+    }
+    return prev.replace(currentRegexResult[0], replacementValue);
+  }, textObject.text);
+  newTextObject.text = newText;
+  return newTextObject;
+};
+
+/** Takes a base pdf, a templateData object that tells us where to put the texts,
+ * and a jsonData object from which we import the values. */
```
dannil76, line 51: then tell me this with a descriptive func name

services/pdf-ms/helpers/pdf.ts
```diff
@@ -0,0 +1,72 @@
...
+export const modifyPdf = async (
+  pdfBuffer: Buffer,
+  templateData: TemplateData,
+  jsonData: Record<string, any>
```
dannil76, line 55: noice word alert

services/pdf-ms/helpers/pdf.ts
```diff
@@ -0,0 +1,72 @@
...
+) => {
+  const pdfDoc = await PDFDocument.load(pdfBuffer);
+  const pages = pdfDoc.getPages();
+  const fonts = await loadFonts(pdfDoc);
+
+  const replacedTextObjects = templateData.textObjects.map(textObject =>
+    replaceTextInTextObject(textObject, jsonData)
+  );
+
```
dannil76, line 64: noice word alert

services/pdf-ms/serverless.yml
```diff
@@ -0,0 +1,86 @@
...
+    resourcesStage: ${self:custom.resourcesStage}
+
+  iamRoleStatements:
+    - Effect: Allow
+      Action:
+        - SSM:GetParameter
+      Resource: "*"
+    - Effect: Allow
+      Action:
+        - dynamodb:ListStreams
```
dannil76, line 44: this is only valid if you use event stream. in this case you'r using event bridge. should be removed

services/pdf-ms/serverless.yml
```diff
@@ -0,0 +1,86 @@
...
+        - Fn::ImportValue: ${self:custom.resourcesStage}-PdfTemplatesBucketArn
+        - Fn::Join:
+            - ""
+            - - Fn::ImportValue: ${self:custom.resourcesStage}-PdfTemplatesBucketArn
+              - "/*"
+
+    - Effect: Allow
+      Action:
+        - dynamodb:Query
+        - dynamodb:Scan
```
dannil76, line 63: do you need Query and Scan permissions?

services/pdf-ms/serverless.yml
```diff
@@ -0,0 +1,86 @@
...
+        - dynamodb:GetItem
+      Resource:
+        - Fn::ImportValue: ${self:custom.resourcesStage}-FormsTableArn
+
+functions:
+  addPdfToCase:
+    handler: lambdas/addPdfToCase.main
```
dannil76, line 70: the name does not say that the lambda actually generates pdf by processing application responses

services/pdf-ms/serverless.yml
```diff
@@ -0,0 +1,86 @@
...
+  resourcesStages:
+    prod: prod
+    dev: dev
+  resourcesStage: ${self:custom.resourcesStages.${self:custom.stage}, self:custom.resourcesStages.dev}
+  webpack:
+    webpackConfig: webpack.config.js
+    includeModules:
+      forceExclude:
+        - aws-sdk
+    packager: "yarn"
```
dannil76, line 18: the serverless-bundle plugin has this coverd? see comments above

services/pdf-ms/templateExample/ekb-template.json
```diff
@@ -0,0 +1,27 @@
+{
+    "numberOfPages": 1,
+    "defaultFontSize": 7,
+    "textObjects": [
```
dannil76, line 4: noice word alert

services/pdf-ms/templateExample/ekb-template.json
```diff
@@ -0,0 +1,27 @@
...
+        {
+            "page": 0,
+            "x": 400,
+            "y": 610,
+            "fontSize": 7,
+            "text": "{{personalInfo.firstName}}"
```
dannil76, line 10: noice word alert

services/pdf-ms/templateExample/ekb-template.json
```diff
@@ -0,0 +1,27 @@
...
+        },
+        {
+            "page": 0,
+            "x": 400,
+            "y": 598,
+            "fontSize": 7,
+            "text": "{{personalInfo.phone}}"
```
dannil76, line 17: noice word alert

services/pdf-ms/templateExample/ekb-template.json
```diff
@@ -0,0 +1,27 @@
...
+        },
+        {
+            "page": 0,
+            "x": 400,
+            "y": 586,
+            "fontSize": 7,
+            "text": "{{personalInfo.email}}"
```
dannil76, line 24: noice word alert

services/pdf-ms/templateExample/ekb-template.json
```diff
@@ -0,0 +1,27 @@
...
+        }
+    ]
+}
```
dannil76, line 27: missing newline at end of file

services/pdf-ms/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "include": ["./**/*.ts"],
+    "exclude": [
+      "node_modules/**/*",
+      ".serverless/**/*",
+      ".webpack/**/*",
+      "_warmup/**/*",
+      ".vscode/**/*"
+    ]
+  }
```
dannil76, line 25: missing newline at end of file

dannil76 (changes requested):
a lot to look at. think we should take a look at it together to reason our way to perfection

dannil76 (changes requested):
a lot to look at. think we should take a look at it together to reason our way to perfection

jwinding:
That's a lot of comments. Went through it and a lot of the complaints about naming I think I could fix. Some less than ideal naming remains, but for those I don't directly know good names, so we can discuss later. 


Then the main thing is the serverless setup, and using serverless-bundle vs serverless-webpack. Here, I can't say that I agree that serverless-webpack is outdated, because while serverless-bundle is nice when it works, it frequently just doesn't. You are stuck with their webpack setup, which seems to not work properly with many (larger) libraries: it seems like it doesn't upload all the required code to the lambda, so I ran into runtime errors during pdf generation. It's possible that there's options for making it work, but if so I do not know how, so to just get something working it was faster to use webpack-bundle so that you could setup the webpack as required. 

Loading some libraries from root instead of having them in the local npm folder is also something that still needs fixing, but this depends on what we arrive at regarding bundle vs webpack, so to not waste time I'm leaving it until we decide something there.

jwinding:
Making some further changes: 

- added loading of the personal information of the user from the user database, since we in the pdf want to write out things like personalNumber and name. 
- Change so that the names of template files are dynamically loaded from a file in the bucket, so when adding/changing some form template, one can just update that file instead of having to redeploy the service.
- some more small naming improvements

jwinding:
Further changes:
- added logic to compare values to last case of the same type, and mark the changes in red in the form
generated pdf

For this to work, I add a global secondary index to the cases db, which can be found in [this PR on the resource repo](https://github.com/helsingborg-stad/helsingborg-io-sls-resources/pull/40), which lets me find relevant cases, and then I take the latest such which is submitted, and compare each value to see what is new and what's changed. Right now we treat those as the same thing and just mark them in red.

services/pdf-ms/helpers/case.ts
```diff
@@ -0,0 +1,55 @@
+import to from 'await-to-js';
+
+import config from '../../../config';
+import * as dynamoDb from '../../../libs/dynamoDb';
+import { Case, AnswerObject } from './types';
+
+/** Get all cases for a given person and formId */
+export const getRelevantCases = async (personalNumber: string, formId: string) => {
```
dannil76, line 8: why not name the fn to something that matches your comment? then delete comment

services/pdf-ms/helpers/case.ts
```diff
@@ -0,0 +1,55 @@
...
+  }>(dynamoDb.call('query', params));
+  if (error) {
+    console.error(error);
+    return undefined;
+  }
+  return casesResponse.Items;
+};
+
+/** compare two cases, returning lists of new and changed values */
+export const compareCases = (
```
dannil76, line 37: getNewAndChangedCases?

services/pdf-ms/helpers/case.ts
```diff
@@ -0,0 +1,55 @@
...
+  currentCase: Case & {
+    answersArray: AnswerObject[];
+  },
+  oldCase: { answers: AnswerObject[] }
+) => {
+  const changedValues: string[] = [];
+  const newValues: string[] = [];
+
+  currentCase.answersArray.forEach(answer => {
```
dannil76, line 46: why forEach?

services/pdf-ms/helpers/pdf.ts
```diff
@@ -50,31 +50,40 @@ const replaceTextInTextObject = (textObject: TextObject, json: Record<string, an
   const templateStrings = [...textObject.text.matchAll(regex)];
   const newTextObject: TextObject = { ...textObject };
 
-  const newText = templateStrings.reduce((prev, currentRegexResult) => {
+  const newText = templateStrings.reduce((previous, currentRegexResult) => {
     const replacement = getPropertyFromDottedString(json, currentRegexResult[1].trim());
```
dannil76, line 54: what is [1]?

services/pdf-ms/helpers/pdf.ts
```diff
@@ -50,31 +50,40 @@ const replaceTextInTextObject = (textObject: TextObject, json: Record<string, an
...
     if (replacement === undefined || replacement === 'undefined') {
-      return prev.replace(currentRegexResult[0], '');
+      return previous.replace(currentRegexResult[0], '');
```
dannil76, line 56: what is [0]?

services/pdf-ms/helpers/pdf.ts
```diff
@@ -50,31 +50,40 @@ const replaceTextInTextObject = (textObject: TextObject, json: Record<string, an
...
-  const replacedTextObjects = template.texts.map(textObject =>
-    replaceTextInTextObject(textObject, jsonData)
-  );
+  const replacedTextObjects = template.texts.map(textObject => {
+    const replacedText = replaceTextInTextObject(textObject, json);
+    if (
+      replacedText.valueId &&
+      (newValues.includes(replacedText.valueId) || changedValues.includes(replacedText.valueId))
+    ) {
+      replacedText.color = rgb(174 / 255, 11 / 255, 5 / 255);
```
dannil76, line 83: const?

services/pdf-ms/serverless.yml
```diff
@@ -44,6 +44,21 @@ provider:
...
+        - dynamodb:Query
+      Resource:
+        - Fn::Join:
+          - ""
+          - - Fn::ImportValue: ${self:custom.resourcesStage}-CasesTableArn
+            - "/index/*"
+    - Effect: Allow
+      Action:
+        - dynamodb:Query
+        - dynamodb:Scan
```
dannil76, line 58: You get the hole document with scan. is this what you want?

services/pdf-ms/templateExample/ekb-löpande.json
```diff
@@ -0,0 +1,480 @@
+{
+  "numberOfpages": 5,
+  "defaultFontSize": 11,
+  "texts":
+  [
+  {
```
dannil76, line 6: missing indentation

services/pdf-ms/templateExample/ekb-löpande.json
```diff
@@ -0,0 +1,480 @@
...
+      "page": 0,
```
dannil76, line 7: more user friendly to start with page nr 1...

services/pdf-ms/templateExample/ekb-löpande.json
```diff
@@ -0,0 +1,480 @@
...
+      "x": 400,
+      "y": 685,
+      "size": 11,
```
dannil76, line 10: size of what?

services/pdf-ms/templateExample/ekb-löpande.json
```diff
@@ -0,0 +1,480 @@
...
+      "text": "{{valuesChanged}}",
+      "valueId": "valuesChanged"
+  },
+  {
+      "page": 0,
+      "x": 270,
+      "y": 648,
```
dannil76, line 17: maybe left, top

services/pdf-ms/templateExample/ekb-löpande.json
```diff
@@ -0,0 +1,480 @@
...
+      "maxWidthInChars": 50,
+      "valueId": "otherMessage"
+  },
+  {
+      "page": 0,
+      "x": 400,
+      "y": 458,
+      "size": 11,
+      "text": "{{user.personalNumber}}",
+      "valueId": "user.personalNumber"
```
dannil76, line 29: what do this do?

dannil76 (changes requested):

services/pdf-ms/helpers/pdf.ts
jwinding: So these results come from the [matchAll function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/matchAll), which finds all regex matches and returns an array of results, where each result has the form of another array of the form [ fullMatch, capturingGroup1, capturingGroup2, ...]. So in this case, I just have one capturing group in my regex, so [1] gets that, while [0] gets the full expression, including the {{ }}.

services/pdf-ms/templateExample/ekb-löpande.json
jwinding: Well, this is for programmers, and we should know that indices starts at zero, no?

services/pdf-ms/helpers/case.ts
jwinding: Why not? In general I do think I find forEach cleaner than for loops, but I don't think it really matters to much which one is used.

services/pdf-ms/helpers/pdf.ts
TeddyAndersson: Suggestion here is to destruct the array values, might be easier to understand then.
```const [ fullMatch, capturingGroup1, capturingGroup2, ...] = currentRegexResult```

File: services/pdf-ms/helpers/pdf.ts (hunk 30 above)
Comment (TeddyAndersson) [reply services/pdf-ms/helpers/pdf.ts:83]: the variabel replacedText is an object, and we do only change properties in the object. Therefore using const here is fine for me. If we where to write over the value of replacedText then a let would be the way to go.

File: services/pdf-ms/helpers/types.ts [hunk 37]
```diff
@@ -0,0 +1,40 @@
...
+  SK: string;
+  status: 'ongoing' | 'submitted';
+  provider: string;
+  expirationTime: string;
+  createdAt: string;
+  updatedAt: string;
+  answers: Record<string, any>;
+  details?: Record<string, any>;
+  pdf?: string | Buffer;
+  pdfGenerated?: boolean;
```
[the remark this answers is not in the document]
TeddyAndersson: instead of pdfIsGenerated i suggest: isPdfGenerated.

services/pdf-ms/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+This services uses a library that lets us render new text on top of an existing pdf, which is how the 'template system' works. The template data consists of this base pdf, and another data object: 
+```typescript
+export interface TemplateData {
+  numberOfPages: number;
+  defaultFontSize: number;
+  defaultFont?: Font;
+  date?: string;
+
+  textObjects: TextObject[];
```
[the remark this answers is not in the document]
TeddyAndersson: @jwinding how would you describe the textObject? ie whats it's purpose. A out of the blue suggestion here might be TemplateTextList or just TextList or something.

services/pdf-ms/README.md
```diff
@@ -0,0 +1,46 @@
...
+}
+```
+For each string that we want to render, we have to specify a TextObject, which has the following shape: 
+```typescript
+interface TextObject {
```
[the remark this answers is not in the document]
TeddyAndersson: Not sure if Object is a reserved type. Either way i rather have a more detailed name for the type than just Object if thats what your suggesting @dannil76

services/pdf-ms/templateExample/ekb-löpande.json
TeddyAndersson: Maybe change the property to pageIndex instead @jwinding, to avoid confusion (y)

services/pdf-ms/helpers/case.ts
TeddyAndersson: forEach is perfectly fine and in my opinon easier to read as-well (y)

services/pdf-ms/helpers/pdf.ts
jwinding: yeah, sounds like a good idea for readability, I'll do that.

services/pdf-ms/templateExample/ekb-löpande.json
jwinding: The text. But I'll change it back to fontSize to be more explicit.

services/pdf-ms/templateExample/ekb-löpande.json
jwinding: It is used to check if the value has changed since last time or not, and if so, change it to have a red color. Unsure where to properly document this.

services/pdf-ms/templateExample/ekb-löpande.json
jwinding: I think x,y is pretty clear, and probably more standard than anything else for giving coordinates? And it's what the pdf library uses.

services/pdf-ms/helpers/case.ts
jwinding: Renamed to getNewAndChangedValues, which still probably isn't perfect, but feels good enough.

services/pdf-ms/README.md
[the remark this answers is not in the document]
dannil76: > Not sure if Object is a reserved type. Either way i rather have a more detailed name for the type than just Object if thats what your suggesting @dannil76

I'm refering to the term Object in TextObject. Don't prefix!
Words like Data, Object, Array etc. is not telling me shit... Instead tell me what to expect in simple, clear english!

services/pdf-ms/helpers/pdf.ts
dannil76: > the variabel replacedText is an object, and we do only change properties in the object. Therefore using const here is fine for me. If we where to write over the value of replacedText then a let would be the way to go.

I mean, put color in some const variable stored in the global const lib file

services/pdf-ms/templateExample/ekb-löpande.json
dannil76: > I think x,y is pretty clear, and probably more standard than anything else for giving coordinates? And it's what the pdf library uses.

standard for who?! An editor | ux-person | developer | me | you

services/pdf-ms/templateExample/ekb-löpande.json
dannil76: Can you name it to that then? last changed ...someting

services/pdf-ms/helpers/case.ts
```diff
@@ -34,7 +33,7 @@ export const getRelevantCases = async (personalNumber: string, formId: string) =
 };
 
 /** compare two cases, returning lists of new and changed values */
```
dannil76, line 35: Delete!

services/pdf-ms/helpers/case.ts
```diff
@@ -4,8 +4,7 @@ import config from '../../../config';
 import * as dynamoDb from '../../../libs/dynamoDb';
 import { Case, AnswerObject } from './types';
 
-/** Get all cases for a given person and formId */
-export const getRelevantCases = async (personalNumber: string, formId: string) => {
+export const getCasesWithFormId = async (personalNumber: string, formId: string) => {
```
dannil76, line 7: getApplicantCasesByFormId?

services/pdf-ms/helpers/pdf.ts
```diff
@@ -22,14 +22,14 @@ const renderTextOnPage = (
...
     const numLines = Math.ceil(text.length / maxWidthInChars);
     for (let line = 0; line < numLines; line++) {
       pdfPage.drawText(text.substring(maxWidthInChars * line, maxWidthInChars * (line + 1)), {
         x,
-        y: y - line * (size * 1.5),
-        size: size || defaultFontSize,
+        y: y - line * (fontSize * 1.5),
+        size: fontSize || defaultFontSize,
         font: fonts[font || 'helvetica'],
         color: color || rgb(0.05, 0.05, 0.05),
```
dannil76, line 34: Put color in const?

services/pdf-ms/lambdas/addPdfToCase.ts
```diff
@@ -26,7 +26,7 @@ export const main = async (event: Record<string, any>) => {
   if (user) caseData.answers.user = user;
 
   // find changed and newly added values, compared to the last case with the same formId.
-  const relevantCases = await getRelevantCases(personalNumber, formId);
+  const relevantCases = await getCasesWithFormId(personalNumber, formId);
```
dannil76, line 29: do i explicit need to know it's relevantCases. Why not just cases?

services/pdf-ms/lambdas/addPdfToCase.ts
```diff
@@ -35,7 +35,7 @@ export const main = async (event: Record<string, any>) => {
       .filter(c => c.SK !== SK)
       .sort((c1, c2) => c2.updatedAt - c1.updatedAt);
```
dannil76: SK, c, c1, c2?!
pronounced names please

services/pdf-ms/lambdas/addPdfToCase.ts
```diff
@@ -35,7 +35,7 @@ export const main = async (event: Record<string, any>) => {
...
     const newestRelevantCase = sortedCases[0];
-    const comparison = compareCases(caseData, newestRelevantCase);
+    const comparison = getNewAndChangedValues(caseData, newestRelevantCase);
```
dannil76, line 38: noice word alert!
Data... RelevantCase....

services/pdf-ms/lambdas/addPdfToCase.ts
```diff
@@ -35,7 +35,7 @@ export const main = async (event: Record<string, any>) => {
...
     changedValues = comparison.changedValues;
     newValues = comparison.newValues;
     caseData.answers.valuesChanged =
```
dannil76, line 41: noice word alert!

services/pdf-ms/templateExample/ekb-löpande.json
```diff
@@ -3,478 +3,478 @@
...
+    {
+        "page": 4,
+        "x": 400,
+        "y": 503,
+        "fontSize": 11,
+        "text": "{{barnomsorg}}",
+        "valueId": "barnomsorg"
+    }
+  ]
 }
```
dannil76, line 480: end of file - missing newline

services/pdf-ms/helpers/caseDataConverter.ts
dannil76: Still can't get this?!

dannil76 (changes requested):

services/pdf-ms/helpers/user.ts
```diff
@@ -0,0 +1,22 @@
...
+import * as dynamoDb from '../../../libs/dynamoDb';
+
+export const getUser = async (personalNumber: string) => {
+  const params = {
+    TableName: config.users.tableName,
+    Key: {
+      personalNumber,
+    },
+  };
+  const [error, userGetResponse] = await to<{ Item: Record<string, any> }>(
```
dannil76, line 13: getUsersDbResponse är trevligare att uttala. Eller?
Ännu bättre är bara usersDbResponse
Passar fint med usersDbResponse.Item

services/pdf-ms/helpers/user.ts
```diff
@@ -0,0 +1,22 @@
...
+    dynamoDb.call('get', params)
+  );
+  if (error) {
+    console.error(error);
+    return undefined;
```
dannil76, line 18: Varför inte return false?

dannil76 (changes requested):

services/pdf-ms/lambdas/addPdfToCase.ts
jwinding: Just case cannot be used since it's a reserved keyword, that's why I'm using caseData. And yeah, maybe Relevant can be removed.

services/pdf-ms/lambdas/addPdfToCase.ts
jwinding: Isn't the letter c perfectly pronouncable though? SK might not be pronouncable, but hey, we've used that as a property name all over the API, so it's not exactly originating here...

Also in my view, for these kind of short inline expressions, it's perfectly readable to have short variable names, since they are only there for this very short expression and not remembered past that.

services/pdf-ms/helpers/pdf.ts
jwinding: What do you mean?

services/pdf-ms/helpers/user.ts
jwinding: Renaming to usersDbResponse.

services/pdf-ms/helpers/user.ts
jwinding: Because it felt better like better typing to have either a json object or undefined, rather than a json or false. Sort of amounts to the same thing though.

services/pdf-ms/lambdas/addPdfToCase.ts
jwinding: The idea was to signal that it's not just any or all cases, but some specific ones.

services/pdf-ms/helpers/pdf.ts
jwinding: Ah, okay, thats what you mean. Then I agree, both color values should be lifted to two consts. But not sure it needs to go to some global lib file, since they are probably only used in this one file, so local constants feels better to me.

services/pdf-ms/serverless.yml
jwinding: I copied this from the usersApi (which I think you wrote 😃), where we allow both Query, Scan and getItem; so I'm unsure. Since I'm using dynamoDb.call('get', params) is getItem all I need? Sounds reasonable, I'll test it, and if so, we should also change this in the usersApi.

jwinding: Checked this, and I only need getItem, so removing the other two. Same change can be made in the usersApi.

services/pdf-ms/helpers/case.ts
```diff
@@ -4,7 +4,7 @@ import config from '../../../config';
 import * as dynamoDb from '../../../libs/dynamoDb';
 import { Case, AnswerObject } from './types';
 
-export const getCasesWithFormId = async (personalNumber: string, formId: string) => {
+export const getApplicantCasesByFormId = async (personalNumber: string, formId: string) => {
```
dannil76, line 7: Nice!

services/pdf-ms/helpers/pdf.ts
```diff
@@ -2,6 +2,9 @@ import { getPropertyFromDottedString } from './objects';
 import { PDFDocument, StandardFonts, rgb, PDFFont, PDFPage } from 'pdf-lib';
 import { TextObject, Font, Template } from './types';
 
+const defaultTextColor = rgb(0.05, 0.05, 0.05);
+const changedTextColor = rgb(174 / 255, 11 / 255, 5 / 255);
```
dannil76, line 6: Kanske petigt men. konstanter får gärna skrivas med versaler för att just belysa att de är konstanter.... Je..

dannil76 (commented):
Närmar sig perfektion...

dannil76:
ser att du mergat. varför?

PR #136, opened 2021-01-15
feat(collect.js): added support for issuing authorization code

TeddyAndersson:

Fix the imageUploader component, so that it is in usable shape and adheres to the Figma design. 

The auth service have been updated with a new flow for obtaining tokens. In order to obtain a token upon login on the auth service the client needs to obtain an authorization code. Since we do not have a designated endpoint for obtaining a authorization code, this will need to be done when the bankid collect request has the status completed.

## Explain your solution

When the bankID collect request returns the status of complete we generate a jwt token that is signed with a secret (AuthorizationCodeSecret) stored in AWS secrets manager. 

When the token is signed we pass generated it in the bankID collect response on the key authorizationCode.

## How to test the changes?

1. checkout this branch.
2. deploy the bankid service.
3. make sure you have all the secrets configured in AWS secrets (can be found in the config.js)
4. start a bankid authentication process (/bankid/auth).
5. request the bankid collect endpoint (/bankid/collect)
6. when the bankid collect response returns completed, make sure it includes an authorizationCode in the response body attributes.

### DIFF
--- a/services/bankid-api/lambdas/collect.js
+++ b/services/bankid-api/lambdas/collect.js
@@ -37,8 +37,10 @@ export const main = async event => {
       'bankId.collect'
     );
 
+    const personalNumber = bankIdCollectResponse.data.completionData.user.personalNumber;
+
     const [generateAuthorizationCodeError, authorizationCode] = await to(
-      generateAuthorizationCode(bankIdCollectResponse.data.completionData.user.personalNumber)
+      generateAuthorizationCode(personalNumber)
     );
 
     if (generateAuthorizationCodeError) {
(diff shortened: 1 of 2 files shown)


dannil76 (approved):
As clean as it gets

dannil76 (approved):
Godis för mina ögon!

jwinding (commented):
So trying to review this, I got my bankid to work after some struggling (I made a mistake in my parameterstore), but since I don't have a correct app version on my phone, I can't really test the function anyways.

But anyhow, the code seems good to me, and from what I understand the handling of bankid is not changed since before, so as long as someone has tested the functionality, you can treat this as an approve. 

A comment that is unrelated to the PR, but could quickly be added to it, is why the bankid endpoints are not behind the api-key? Not a big deal but seems inconsistent since we have it on the /auth/token endpoints.

TeddyAndersson:
This branch have now been tested with BankID and some minor fixes have been added in order for the collect complete function as intended.

jwinding (approved):
Code looks good, can't test it fully because of bankid, but I'll believe that you tested it.

PR #142, opened 2021-02-03
Fix/status objects

jonatanhanson:
Implemented status as objects instead of string.

If you are testing the changes with the app you need to checkout branch: feat/case-overview-handle-periods

### DIFF
--- a/libs/caseStatuses.js
+++ b/libs/caseStatuses.js
@@ -29,41 +29,41 @@ const statuses = [
    * Service: Ekonomiskt bistånd
    */
   {
-    type: 'notStarted.ekb.open',
+    type: 'notStarted.ekb',
     name: 'Öppen',
     description: 'Ansökan är öppen. Du kan nu söka ekonomiskt bistånd för perioden.',
   },
   {
-    type: 'active.ekb.submitted',
+    type: 'active.submitted.ekb',
     name: 'Inskickad',
     description:
       'Ansökan är inskickad. Du kommer att få besked om ansökan när din handläggare har granskat och bedömt den.',
   },
   {
-    type: 'active.ekb.completionRequired',
+    type: 'active.completionRequired.ekb',
     name: 'Stickprovskontroll',
     description:
       'Du måste komplettera din ansökan med bilder som visar dina utgifter och inkomster. Vi behöver din komplettering inom 4 dagar för att kunna betala ut pengar för perioden.',
   },
   {
-    type: 'closed.ekb.approved',
+    type: 'closed.approved.ekb',
     name: 'Godkänd',
     description: 'Din ansökan är godkänd. Pengarna sätts in på ditt konto.',
   },
   {
-    type: 'closed.ekb.partiallyApproved',
+    type: 'closed.partiallyApproved.ekb',
     name: 'Delvis godkänd',
     description:
       'Delar av din ansökan är godkänd, men några av de utgifter du sökt för får du inte bistånd för. Pengarna för godkända utgifter sätts in på ditt konto.',
   },
   {
-    type: 'closed.ekb.rejected',
+    type: 'closed.rejected.ekb',
     name: 'Avslagen',
     description:
       'Din ansökan är inte godkänd och du kommer inte att få någon utbetalning. Vill du överklaga beslutet lämnar du en skriftlig motivering med e-post eller brev till din handläggare.',
   },
   {
-    type: 'closed.ekb.completionRejected',
+    type: 'closed.completionRejected.ekb',
     name: 'Avslagen',
     description:
       'Din ansökan är inte godkänd eftersom vi saknar stickprov för perioden. Därför kan vi inte gå vidare och godkänna din ansökan.',
--- a/libs/caseStatuses.js
+++ b/libs/caseStatuses.js
@@ -5,18 +5,18 @@ const statuses = [
     description: 'Ansökan är ej påbörjad.',
   },
   {
-    type: 'active.ongoing',
+    type: 'active:ongoing',
     name: 'Pågående',
     description:
       'Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade.',
   },
   {
-    type: 'active.submitted',
+    type: 'active:submitted',
     name: 'Inskickad',
     description: 'Ansökan är inskickad.',
   },
   {
-    type: 'active.processing',
+    type: 'active:processing',
     name: 'Ansökan behandlas',
     description: 'Ditt ärende är mottaget och bearbetas.',
   },
@@ -29,41 +29,41 @@ const statuses = [
    * Service: Ekonomiskt bistånd
    */
   {
-    type: 'notStarted.ekb',
+    type: 'notStarted:ekb',
     name: 'Öppen',
     description: 'Ansökan är öppen. Du kan nu söka ekonomiskt bistånd för perioden.',
   },
   {
-    type: 'active.submitted.ekb',
+    type: 'active:submitted:ekb',
     name: 'Inskickad',
     description:
       'Ansökan är inskickad. Du kommer att få besked om ansökan när din handläggare har granskat och bedömt den.',
   },
   {
-    type: 'active.completionRequired.ekb',
+    type: 'active:completionRequired:ekb',
     name: 'Stickprovskontroll',
     description:
       'Du måste komplettera din ansökan med bilder som visar dina utgifter och inkomster. Vi behöver din komplettering inom 4 dagar för att kunna betala ut pengar för perioden.',
   },
   {
-    type: 'closed.approved.ekb',
+    type: 'closed:approved:ekb',
     name: 'Godkänd',
     description: 'Din ansökan är godkänd. Pengarna sätts in på ditt konto.',
   },
   {
-    type: 'closed.partiallyApproved.ekb',
+    type: 'closed:partiallyApproved:ekb',
     name: 'Delvis godkänd',
     description:
       'Delar av din ansökan är godkänd, men några av de utgifter du sökt för får du inte bistånd för. Pengarna för godkända utgifter sätts in på ditt konto.',
   },
   {
-    type: 'closed.rejected.ekb',
+    type: 'closed:rejected:ekb',
     name: 'Avslagen',
     description:
       'Din ansökan är inte godkänd och du kommer inte att få någon utbetalning. Vill du överklaga beslutet lämnar du en skriftlig motivering med e-post eller brev till din handläggare.',
   },
   {
-    type: 'closed.completionRejected.ekb',
+    type: 'closed:completionRejected:ekb',
     name: 'Avslagen',
     description:
       'Din ansökan är inte godkänd eftersom vi saknar stickprov för perioden. Därför kan vi inte gå vidare och godkänna din ansökan.',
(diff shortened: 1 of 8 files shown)


libs/caseStatuses.js
```diff
@@ -0,0 +1,75 @@
...
+    type: 'closed.ekb.completionRejected',
+    name: 'Avslagen',
+    description:
+      'Din ansökan är inte godkänd eftersom vi saknar stickprov för perioden. Därför kan vi inte gå vidare och godkänna din ansökan.',
+  },
+];
+
+export default statuses;
+
+export const getStatusByType = type => statuses.find(status => status.type === type);
```
TeddyAndersson, line 75: normal exports should be defined before default export 👍

TeddyAndersson (commented):
Just found one minor thing in the code changes, will need to test further. Since there are a number of cases :)

jwinding:
> I think i have changed everywhere we handle statuses in the code but i don't know how this line should look like:
> https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/7b89f0a0aaa45ea3948d1d252f93b573e66c9de9/services/pdf-ms/serverless.yml#L95
> 
> 
> So if someone could fix that it would be much appreciated.

That line is testing for the pattern that we want to trigger the pdf generation. So it has to be replaced with another  pattern that matches your new status, more precisely against the type inside the status object. I don't know the correct syntax for this of the top of my head, but I'll try and find out.

dannil76:
> > I think i have changed everywhere we handle statuses in the code but i don't know how this line should look like:
> > https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/7b89f0a0aaa45ea3948d1d252f93b573e66c9de9/services/pdf-ms/serverless.yml#L95
> > 
> > So if someone could fix that it would be much appreciated.
> 
> That line is testing for the pattern that we want to trigger the pdf generation. So it has to be replaced with another pattern that matches your new status, more precisely against the type inside the status object. I don't know the correct syntax for this of the top of my head, but I'll try and find out.

Fixar detta då jag känner till syntaxen... :)

jonatanhanson:
> > I think i have changed everywhere we handle statuses in the code but i don't know how this line should look like:
> > https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/7b89f0a0aaa45ea3948d1d252f93b573e66c9de9/services/pdf-ms/serverless.yml#L95
> > 
> > So if someone could fix that it would be much appreciated.
> 
> That line is testing for the pattern that we want to trigger the pdf generation. So it has to be replaced with another pattern that matches your new status, more precisely against the type inside the status object. I don't know the correct syntax for this of the top of my head, but I'll try and find out.

Unfortunately we cannot filter events by suffixes or regex so the best i could do was adding the submitted types to an array. 
I dont know how to test the PDF service if this actually works. Can someone who knows test it? 😅 
@jwinding @TeddyAndersson

libs/caseStatuses.js
jonatanhanson: fixed

TeddyAndersson:
> > > I think i have changed everywhere we handle statuses in the code but i don't know how this line should look like:
> > > https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/7b89f0a0aaa45ea3948d1d252f93b573e66c9de9/services/pdf-ms/serverless.yml#L95
> > > 
> > > So if someone could fix that it would be much appreciated.
> > 
> > 
> > That line is testing for the pattern that we want to trigger the pdf generation. So it has to be replaced with another pattern that matches your new status, more precisely against the type inside the status object. I don't know the correct syntax for this of the top of my head, but I'll try and find out.
> 
> Unfortunately we cannot filter events by suffixes or regex so the best i could do was adding the submitted types to an array.
> I dont know how to test the PDF service if this actually works. Can someone who knows test it? 😅
> @jwinding @TeddyAndersson



> > > I think i have changed everywhere we handle statuses in the code but i don't know how this line should look like:
> > > https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/7b89f0a0aaa45ea3948d1d252f93b573e66c9de9/services/pdf-ms/serverless.yml#L95
> > > 
> > > So if someone could fix that it would be much appreciated.
> > 
> > 
> > That line is testing for the pattern that we want to trigger the pdf generation. So it has to be replaced with another pattern that matches your new status, more precisely against the type inside the status object. I don't know the correct syntax for this of the top of my head, but I'll try and find out.
> 
> Unfortunately we cannot filter events by suffixes or regex so the best i could do was adding the submitted types to an array.
> I dont know how to test the PDF service if this actually works. Can someone who knows test it? 😅
> @jwinding @TeddyAndersson

I can test to see if the lambda triggers on the event pattern (y)

jonatanhanson:
Starting to think if it would be better to name types like this instead "active.submitted.ekb" instead of "active.ekb.submitted", so we can use prefixes for example "active.submitted" for matching all submitted cases. Or what do you guys think? 
@TeddyAndersson @dannil76 @jwinding @zilaei

jwinding:
> Starting to think if it would be better to name types like this "active.submitted.ekb" instead of "active.ekb.submitted", so we can use prefixes for example "active.submitted" for matching all submitted cases. Or what do you guys think?
> @TeddyAndersson @dannil76 @jwinding @zilaei

I agree with this, I had the same thought when looking at your statuses actually. It feels better to add the type of case at the end, that way it gets more specific as you progress, and as you say, you can match prefixes. 

Oh, and I can also test if the pdf lambda triggers, if need be.

TeddyAndersson:
So tried to trigger the lambda through the event pattern. By updating the status on a item in the case table. But the lambda does not react to the event passed through the eventbridge. Trying to figure out how the object/map is defined in the NewImage that the event pattern is relaying on. My guess is that the pattern does not trigger since the value is not correct.


{
...
detail: 
  dynamodb:
    NewImage: 
      status: 
        type: ["active.submitted", "active.ekb.submitted"] // my guess is that it does not know how to read the type value in the map/object data.
}

Specifying that it should match type as a String (S) does not work either. Need to read the docs to see what i can find.
{
...
detail: 
  dynamodb:
    NewImage: 
      status: 
        type:
          S: ["active.submitted", "active.ekb.submitted"]

jonatanhanson:
> So tried to trigger the lambda through the event pattern. By updating the status on a item in the case table. But the lambda does not react to the event passed through the eventbridge. Trying to figure out how the object/map is defined in the NewImage that the event pattern is relaying on. My guess is that the pattern does not trigger since the value is not correct.
> 
> ```
> {
> ...
> detail: 
>   dynamodb:
>     NewImage: 
>       status: 
>         type: ["active.submitted", "active.ekb.submitted"] // my guess is that it does not know how to read the type value in the map/object data.
> }
> ```
> 
> Specifying that it should match type as a String (S) does not work either. Need to read the docs to see what i can find.
> 
> ```
> {
> ...
> detail: 
>   dynamodb:
>     NewImage: 
>       status: 
>         type:
>           S: ["active.submitted", "active.ekb.submitted"]
>  }```
> ```

I changed it to this instead, hope it works 🤞 
`
S: [{"prefix": "active.submitted"}]
`

jwinding:
Okay, solved it! 

> Specifying that it should match type as a String (S) does not work either. Need to read the docs to see what i can find.
> 
> ```
> {
> ...
> detail: 
>   dynamodb:
>     NewImage: 
>       status: 
>         type:
>           S: ["active.submitted", "active.ekb.submitted"]
>  }```
> ```

I also tried this, and it didn't work. Then I logged the event from the stream-handler, and the status part now looks like this: 

```
"status": {
                "M": {
                    "name": {
                        "S": "Inskickad"
                    },
                    "description": {
                        "S": "Ansökan är inskickad. Du kommer att få besked om ansökan när din handläggare har granskat och bedömt den."
                    },
                    "type": {
                        "S": "active.submitted"
                    }
                }
            }
```
so that's what we have to match. The tricky thing is that dynamoDB inserts its datatypes into the structure, which only effects the part coming from the record, not the outer parts that are added by the stream lambda. And the pattern have to include these correctly, or it won't work . So the pattern that finally worked for me looks like
```
            detail:
              dynamodb:
                NewImage:
                    status:
                      M:
                        type: 
                          S: ["active.submitted",  "active.ekb.submitted"]
```
This can presumably be updated to match prefixes, but it worked for me when I sent a status with type: active.submitted at least.

TeddyAndersson:
Came to the same conclusion as jacob. Do you push the changes, @jwinding? Or should i? ☺️

TeddyAndersson:
Adding prefix to the type parttern, seems like a good idea. Will need to change the dotnotation on types aswell. 

A personal suggestion here is to use another seperator in order to not mix type values with js object dotnotation.

Semicolon, colon or a hash is my suggestion

active;submitted;ekb
active:submitted:ekb
active#submitted#ekb

jwinding:
Pushed the changes.

A thing to note here when deploying to dev, is that we will need to clear all old cases, since the updates to the pdf-ms changes the way it looks for statuses, so it breaks if it hits a case with the old status as just a string. I cleared them on my own environment, and then checked that the pdf was generated correctly, which it was.

jonatanhanson:
> Adding prefix to the type parttern, seems like a good idea. Will need to change the dotnotation on types aswell ie (active.submitted.ekb).
> 
> A personal suggestion here is to use another seperator in order to not mix type values with js object dotnotation.
> 
> Semicolon, colon or a hash is my suggestion
> 
> active;submitted;ekb
> active:submitted:ekb
> active#submitted#ekb

I like dots as separator because of its similarity with js dot notation 😄 
But if thats confusing I vote for colon: "active:submitted:ekb"

TeddyAndersson:
> > Adding prefix to the type parttern, seems like a good idea. Will need to change the dotnotation on types aswell ie (active.submitted.ekb).
> > 
> > A personal suggestion here is to use another seperator in order to not mix type values with js object dotnotation.
> > 
> > Semicolon, colon or a hash is my suggestion
> > 
> > active;submitted;ekb
> > active:submitted:ekb
> > active#submitted#ekb
> 
> I like dots as separator because of its similarity with js dot notation 😄 
> But if thats confusing I vote for colon: "active:submitted:ekb"

It's personal preference for me, and not a major issue ☺️

Colon as a seperator instead would be appreciated 👏

TeddyAndersson:
Have tested the following lambdas that are affected by this PR:
addPdfToCase,
createVivaCase,
createCase

Can't find any issues in my environment.

TeddyAndersson (approved):
Good work, seems to work as expected (y)

TeddyAndersson:
> Have tested the following lambdas that are affected by this PR:
> addPdfToCase,
> createVivaCase,
> createCase
> 
> Can't find any issues in my environment.

TeddyAndersson:
Accidentally closed the PR, sorry for that 😅

TeddyAndersson:
> Pushed the changes.
> 
> A thing to note here when deploying to dev, is that we will need to clear all old cases, since the updates to the pdf-ms changes the way it looks for statuses, so it breaks if it hits a case with the old status as just a string. I cleared them on my own environment, and then checked that the pdf was generated correctly, which it was.

Will add a task for this in the sprint so that we do not forget :)

jonatanhanson:
Changed type separators to colons and rebased.

jwinding (approved):
Tested the createCase and addPdfToCase lambdas in my personal environment, and found no issues!

PR #143, opened 2021-02-04
fix(viva-ms): updated status prefix for viva statuses from :ekb to :viva

TeddyAndersson:
Changed prefix on status types ending with ekb to viva.

Since ekb is a acronym and is not used anywhere else in the code base, this can be hard to understand. Viva on the other hand is used as a case provider and is mentioned throughout the code base.

These changes hade it's origin from a conversation with @jonatanhanson. We both agreed that this was a better naming convention for the status types 👍 

## Explain your solution
All status types that ended with :ekb does now end with :viva. Lambdas that relies on setting and getting these statuses have also been updated.

### DIFF
--- a/libs/caseStatuses.js
+++ b/libs/caseStatuses.js
@@ -29,41 +29,41 @@ const statuses = [
    * Service: Ekonomiskt bistånd
    */
   {
-    type: 'notStarted:ekb',
+    type: 'notStarted:viva',
     name: 'Öppen',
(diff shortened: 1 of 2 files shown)


jonatanhanson (approved):

TeddyAndersson:
Think i covered all the previous :ekb statuses even in lambda. Can't find any decalaration of ekb atleast :)

PR #149, opened 2021-02-09
Feat/deploy pipelines

yo-l1982:
## What does this do.

Adds configuration changes to SLS to handle additional test/staging environments.
Adds deploy script.
Includes CFN templates.
Includes some readme about the deploys.

## How to review

Please check my js-code mainly.
The templates and scripts is kind of hard to test if not on a live environment but if you want to I can go through them


.deploy/helpers/diff.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  // Backtrack the path to find the serverless file to get the dirnames of services we want to deploy.
+  const files = [];
+  servicesDiff.forEach(diff => {
+    let diffPath = diff;
+    while (diffPath !== '.') {
+      const serviceFile = `${config.codeBuildPath}/${diffPath}/serverless.yml`;
+      if (fs.existsSync(serviceFile)) {
+        if (!files.includes(serviceFile)) {
```
jwinding, line 54: maybe cleaner to combine these if statements,
```
if (fs.existsSync(serviceFile) && !files.include(serviceFile) { ... } 
```

.deploy/helpers/rollback.js
```diff
@@ -0,0 +1,62 @@
...
+import process from 'process';
+
+import config from '../config.js';
+
+/**
+ * Rollback to previous deployed versions.
+ * @param {array} deployedServices - List of deployed services to rollback.
+ * @return {void}
+ */
+export default deployedServices => {
```
jwinding, line 12: Same style comment again about giving functions names.

.deploy/helpers/rollback.js
```diff
@@ -0,0 +1,62 @@
...
+  for (const deployedService of deployedServices) {
+    let output = '';
+    process.chdir(deployedService);
+
+    // List previous deployed versions of this service with deploy list command.
+    console.log(`Started rollback in: ${deployedService}`);
+    try {
+      output = childProcess.execSync(`${config.listCommand}`).toString();
+    } catch (_error) {
+      console.log('Error listing deployed releases!');
```
jwinding, line 27: Not sure about this, perhaps error messages are already logged as a result of the execSync? Otherwise maybe log some information about the error.

jwinding (commented):
So I went through the javascript part and left a few minor style comments. Not exactly easy to understand the full deploy logic, but I think I at least get some idea, and overall the logic in the javascript seems fairly clear. 

I have only read the code, did not test anything, as that seems a bit more involved.

.deploy/helpers/deployOrder.js
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Based on export and imports in serverless templates, sort the list to get exports before any import.
+ * @param {array} files - List of paths to services files.
+ * @return {array} - Sorted list with exports first.
+ */
+export default files => {
+  const services = [];
+  // Collect all import and exports in the serverless files.
+  files.forEach(file => {
```
TeddyAndersson, line 16: Would use the js map utility function here to iterate over the files array
const services = files.map( file => { ...try catch logic } )

.deploy/helpers/deployOrder.js
```diff
@@ -0,0 +1,72 @@
...
+        imports: imports,
+        exports: jsonpath.query(doc, '$.resources.Outputs.*.Export.Name'),
+      });
+    } catch (e) {
+      console.log(e);
+    }
+  });
+
+  // Sort services according to import and exports.
+  services.sort(sortDeploys);
```
TeddyAndersson, line 33: Preferably the sortDeploy function should be defined before usage here. Also it's a good idea to have the default export on the last line of the file (y)

.deploy/helpers/deployOrder.js
```diff
@@ -0,0 +1,72 @@
...
+  return sortedFiles;
+};
+
+/**
+ * Sorting callback function for import/export sorting.
+ * @param {string} first - First compare string
+ * @param {string} second - Second compare string
+ * @return {int} - Sort order int.
+ */
+const sortDeploys = (first, second) => {
```
TeddyAndersson, line 48: Suggestion here is to have a better naming convention for the first and second arguments. Perhaps firstCompareString and secondCompareString? The comment tells me what the arguments is, but would be better to have more detailed argument naming, then we do not need the comment (y)

TeddyAndersson (changes requested):
Some readability and code style comments. Hard to follow the logic with a huge PR like this.

.deploy/helpers/diff.js
yo-l1982: I want to break out of the while loop only if the second if statement is true.
Combining it breaks the functionality so I think ill skip this style fix, or if you got a better suggestion I can use Ill do it.

yo-l1982:
@TeddyAndersson @jwinding Fixed style stuff and found a small bug that got fixed as well.

.deploy/helpers/diff.js
TeddyAndersson: How does it break? from what i can se both of the checks needs to be true in order to break out of the loop?

TeddyAndersson: Just out of curiosity from my point of view

TeddyAndersson (approved):
From a code perspective i think this is fine (y)
Good work @yo-l1982 👯

.deploy/helpers/diff.js
jwinding: Yeah, I don't really understand either, if only one of them is true, then nothing happens, it seems.

jwinding (approved):
Agree with Teddy, now the javascript part looks fine.

.deploy/helpers/diff.js
yo-l1982: Yeah dont know what I thought here, it broke when I tested but thats probably something else.
Ill try this again.

yo-l1982: I merged the ifs, must have had some typo first try.

PR #151, opened 2021-02-10
Fix/update case new data struct

dannil76:
Vad är ändrat?
Skapat förutsättningar för att endpoint PUT /cases/<id> ska kunna uppdatera stickprovsärende

Lösning:
Ändrat payload: formId till currentFormId
currentFormId är ett krav vid PUT om du uppdaterar answers och/eller currentPosition

currentFormId används för att 'peka' mot rätt form-objekt då ett form sparas med formId som nyckel under forms-objektet

Testa
Checkout branch
Deploy
Skicka PUT /cases/<caseId>
Payload:
`{
  "statusType": "notStarted",
  "currentFormId": "123",
  "currentPosition": {
    "currentMainStep": 10,
    "currentMainStepIndex": 0,
    "index": 0,
    "level": 0
  },
  "answers": [
    {
      "field": {
        "id": "someNewId",
        "tags": [
          "expenses",
          "rent",
          "date"
        ]
      },
      "value": 1614162400000
    }
  ]
}`

### DIFF
--- a/services/cases-api/README.md
+++ b/services/cases-api/README.md
@@ -34,18 +34,35 @@ Request type: GET
 Headers
 Authorization: {user id}
 
-## Example json payload (POST)
+## Example json payload (PUT)
 
 ```
 {
-	"personalNumber": 198602011212,
-	"type": "VIVA_CASE",
-	"data": {
-		"id1": "value12",
-		"id2": "value23",
-		"id3": "value34",
-		"id4": [1323,3566,6443,2345,23321]
-	}
+  "status": {
+    "type": "active:ongoing",
+    "name": "Pågående",
+    "description": "Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade."
+  },
+  "currentFormId": "123",
+  "currentPosition": {
+    "currentMainStep": 10,
+    "currentMainStepIndex": 0,
+    "index": 0,
+    "level": 0
+  },
+  "answers": [
+    {
+      "field": {
+        "id": "someNewId",
+        "tags": [
+          "expenses",
+          "rent",
+          "date"
+        ]
+      },
+      "value": 1614162400000
+    }
+  ]
 }
 ```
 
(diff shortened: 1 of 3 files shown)


jwinding (changes requested):
Code looks good, but while testing I found a thing to improve. If I send a request with some answers, and a currentFormId that does not yet exist in the forms object, I get back an error with code 400 and the message "The document path provided in the update expression is invalid for update". I think the behaviour here should be to create a new entry in the forms object with the new currentFormId, and add the answers there. 

We could also go the route where we say that when the case is created, all the allowed forms are added, and then you are only allowed to enter answers belonging to these forms.

helsingborgsstad:
Task linked: [CU-e0y8z2 updateCase lambda ska stödja ny datastruktur för case data (DynamoDB)](https://app.clickup.com/t/e0y8z2)

jwinding (commented):
After some discussion and clarification, we don't want the updateCase endpoint to be able to add a new form to the forms-object, in which case my previous change request is moot. Instead, I would then want a better error message which makes it clear that we can only update answers on existing forms. Also, documenting this behaviour somewhere is a good idea, like in the Readme.

services/cases-api/README.md
```diff
@@ -34,47 +34,114 @@ Request type: GET
...
-		"id1": "value12",
-		"id2": "value23",
-		"id3": "value34",
-		"id4": [1323,3566,6443,2345,23321]
-	}
+  "status": {
+    "type": "active:ongoing",
+    "name": "Pågående",
+    "description": "Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade."
+  },
```
TeddyAndersson, line 45: A suggestion here is to only submit the statusType, then the lambda can retrieve the additional information saved on the status type. Unless we want to save the status information in the app and on AWS, this would result in two data sources that we need to update instead of one.
statusType: "active:ongoing"

dannil76: Done!

TeddyAndersson:
Another thought here, should we not validate the event body with joi?

dannil76:
agree... add task to backlog

jwinding (changes requested):
Code looks good, but it does not work as is. You need to add an iamRoleStatement in the serverless file,
```
    - Effect: Allow
      Action:
        - dynamodb:Query
      Resource:
        - "Fn::ImportValue": ${self:custom.resourcesStage}-FormsTableArn
  ```
with this it seems to work as intended, tested on my personal environment.

jwinding (approved):

zilaei (approved):

PR #159, opened 2021-02-23
feat(get-attachment): add lambda for downloading an attachment

jwinding:
## What this does:

Adds a lambda for downloading a users attachments, via a get request against /users/me/attachments/{filename}. The response is directly the binary file (which was a little tricky to get working correctly), which should make it easy to consume as you don't have to do any special decoding. 

This is probably useful functionality in general, but right now it's for the ability to resume stickprov.

## Implementation details

The lambda itself is rather straightforward, we just get the file from the S3 bucket, and return a response where the body is the file data in base64 encoded form, together with the minimally required headers and other properties needed. Then two serverless-plugins are used to handle the AWS setup required for binary files. To this end, some configuration is added to the serverless.common file. It's the second one of these plugins that makes the returned response into a binary file. 

## How to test

You'll need to run yarn in the users-api folder, to install the sls plugins. Then deploy the users-api (or just the getAttachment lambda).

Then, check that you have some file (of type jpg, png or pdf) in your attachment S3 bucket, for example let's say you have a file under "201111111111/test.png". Then, with an Authorization token with pnr  201111111111, a get request against "/users/me/attachments/test.png" with the additional header "Accept: image/png" should return the binary png file correctly, and it should show up as an image in Postman/Insomnia. 

(I tried getting rid of the need to send the Accept header, but that seems difficult.)


helsingborgsstad:
Task linked: [CU-et1rw2 Stickprov: API-endpoint för att ladda ner attachments](https://app.clickup.com/t/et1rw2)

services/users-api/lambdas/getAttachment.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable no-console */
```
TeddyAndersson, line 1: remove comment, no console.log declared in the file.

services/users-api/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "Helsingborg Stad",
+  "license": "MIT",
+  "dependencies": {},
+  "devDependencies": {
+    "serverless-apigw-binary": "^0.4.4",
+    "serverless-apigwy-binary": "^1.0.0"
```
TeddyAndersson, line 15: When looking up these plugins on GitHub, the last changes was made 4 and 2 years ago. They do work, but it might cause problem since the repos are inactive. Is this gateway configuration hard to setup without plugins?

TeddyAndersson (changes requested):
Found one minor fix and then i have some concerns about the api-gateway serverless plugins.

services/users-api/lambdas/getAttachment.js
jwinding: fixed.

services/users-api/package.json
jwinding: That's a good observation, I didn't think of that. I found these packages through a rather recent answer on stack overflow. 

To answer the question about setting up the gateway, I don't really know. I tried this at first, there's a gateway setting for allowing certain mime types of binary files through the gateway and so on, and this is described in their documentation, which I followed but couldn't get to work. The lambda would run fine up to the return statement, but then just return the "500: Internal server error"  response, and the CloudWatch logs shows nothing. So there's probably some additional setting that I couldn't find. Whereas these packages made it work (and also presumably helps with automatic deploy, since you don't have to manually change some settings on the gateway). In general, making this work as I wanted was a bit of a pain, so I was just happy when I found something that worked. 
But this is probably a question for Joel or someone else with deeper knowledge of AWS gateways. One can also just look at what these libraries are doing, because they don't seem very big or complicated.


As a further comment, note that there's something strange happening with the serverless-apigwy-binary package. My response contains a base64-encoded string of the file, which then this package somehow converts back to a binary. This is some slightly unsettling magic, but it does work.

TeddyAndersson: Since it works, i don't see any issues with this at the time being. But in the long run this could cause problems. So the question i think we should ask ourself, if this is a technical debt worth adding due to the time constrains we have at the moment. My answer to this is that i think it's worth taking the risk, but we should document the "issue" and tackle it in a future sprint (y)

TeddyAndersson: @yo-l1982 do you have anything to add on this matter?

dannil76 (commented):
Testat att hämta 'hem' en png-bild som ligger på min lokala AWS-miljö.
Använde Postman för att sparat svars-datan på disk. Verkar dock som filen är korrupt när jag vill öppna upp filen?

PR #165, opened 2021-03-08
chore: 🤖 (sls) upgrade serverless.yml

dannil76:
Adapt to new version of serverless and solve serverless deploy deprecation warnings

BREAKING CHANGE: 🧨
Older version of serverless will probably not work

Should not affect the code base

How to test
Have the latest serverless installed - npm i -g serverless
Checkout branch
run sls deploy for every services
test api through app or postman

In some rare cases sls deploy will not update the code base correctly. Then you should run sls remove before sls deploy. This to ensure that the code base actually is the one in your branche

### DIFF
--- a/services/auth/token/serverless.yml
+++ b/services/auth/token/serverless.yml
@@ -11,6 +11,7 @@ package:
   individually: true
 
 provider:
+  lambdaHashingVersion: 20201221
   name: aws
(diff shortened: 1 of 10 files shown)


yo-l1982:
This works like a charm!

yo-l1982 (approved):

PR #169, opened 2021-03-16
Fix/viva api timeout

TeddyAndersson:
- Moved all the viva adapter requests to helper functions
- Removed hashedPersonalNumber from details (payload) in the submitApplicationSuccess event.

Avoid timeout errors in lambda when sending requests to the viva adapter.

Before the configuration of a request to the vada adapter was defined in each lambda. This made each request definition responsible for setting the appropriate timeout time. Now instead we define the request client in a helper function where all the general settings is applied (timeout time, api-key, base-url etc). Each request is then defined in it's own helper function that takes advantage of the client created.

## How to test the changes?
Since these test relies on Viva configurations, the testing is best done in a peer review session with @dannil76 or anyone else that can setup an open application period in viva for a person to apply on.

We want to test the complete flow starting from the point where a user submits a recurring application. Since the api's towards the app is the same, the test can be done by mocking a http post request on /cases/{id}

- [X] AWS on personal environment.


dannil76 (approved):
Tror det funkar...

PR #170, opened 2021-03-18
Sync workflow status rejected, approved, partiallyApproved and completionRequired

zilaei:
Updating case with VIVA statuses rejected, approved,  partiallyApproved and completionRequired.

## Explain your solution
For rejected, approved and  partiallyApproved we check decisionStatus in workflow before setting status for case accordingly.
If several decisions are made, for the same case, we set status partially approved if we find both rejected and approved decisions.

### Test partially approved
**Setup case in AWS:**
```
{
  "answers": [

  ],
  "createdAt": 1611758051108,
  "currentPosition": 3,
  "details": {
    "period": {
      "end": "2021-02-28",
      "start": "2021-02-01"
    },
    "workflowId": "52B708385081D282C125869B0048416D"
  },
  "expirationTime": 1,
  "formId": "0",
  "id": "ac48eedf-819f-4011-bbe8-da6cc2c8c5b4",
  "pdfGenerated": true,
  "PK": "USER#196001198685",
  "provider": "VIVA",
  "SK": "USER#196001198685#CASE#ac48eedf-819f-4011-bbd8-da6cc2c8c5b4",
  "status": {

  },
  "updatedAt": 1611758051108
}
```

**By triggering event from Amazone event bridge:**
Event bus: default
Event source: bankId.collect
Detail type: BankIdCollectComplete
Event detail: { "user": { "personalNumber": "196001198685" } }

### DIFF
--- a/services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
+++ b/services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
@@ -105,17 +105,28 @@ async function syncWorkflowAndStatus(PK, SK, workflow) {
   let UpdateExpression = `SET ${CASE_WORKFLOW_PATH} = :newWorkflow`;
   const ExpressionAttributeValues = { ':newWorkflow': workflow };
   const ExpressionAttributeNames = {};
+  let decisionStatus = 0;
 
-  const typeCode = workflow.decision?.decisions?.decision?.type?.typecode;
+  const decisionList = workflow.decision?.decisions?.decision;
+  if (decisionList === undefined) {
+    // TODO: Return error.
+  }
 
-  if (typeCode === '01') {
+  decisionList.forEach(decision => {
+    const decisionType = decision.typecode;
+    decisionStatus = decisionStatus | parseInt(decisionType, 10);
+  });
+
+  if (decisionStatus === 0) {
+    // TODO: Check if stickprov.
+  }
+
+  if (decisionStatus === 1) {
     ExpressionAttributeValues[':newStatus'] = getStatusByType('closed:approved:viva');
-  } else if (typeCode === '02') {
+  } else if (decisionStatus === 2) {
     ExpressionAttributeValues[':newStatus'] = getStatusByType('closed:rejected:viva');
-  } else if (typeCode === '03') {
+  } else if (decisionStatus === 3) {
     ExpressionAttributeValues[':newStatus'] = getStatusByType('closed:partiallyApproved:viva');
-  } else if (workflow.calculations) {
-    ExpressionAttributeValues[':newStatus'] = getStatusByType('active:processing');
   }
 
   if (ExpressionAttributeValues[':newStatus']) {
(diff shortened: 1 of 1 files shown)


services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
```diff
@@ -13,9 +13,31 @@ export async function main(event) {
   const personalNumber = event.detail.user.personalNumber;
   const PK = `USER#${personalNumber}`;
 
-  const allUserCases = await getAllUserCases(PK);
+  const [getAllUserCasesError, allUserCases] = await to(getAllUserCases(PK));
+  if (getAllUserCasesError) {
+    return console.error('(Viva-ms) DynamoDB query failed', getAllUserCasesError);
+  }
```
TeddyAndersson, line 19: Filter cases by status type active:submitted:viva, to reduce query time.

services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
```diff
@@ -13,9 +13,31 @@ export async function main(event) {
...
+  const userCaseItems = allUserCases.Items;
+  if (userCaseItems === undefined || userCaseItems.length === 0) {
+    return console.error('(Viva-ms) DynamoDB query did not fetch any cases');
+  }
+
+  for (const userCase of userCaseItems) {
+    const workflowId = userCase.details.workflowId;
+
+    const [myPagesError, myPagesResponse] = await to(
+      vivaAdapter.workflow.get({ personalNumber, workflowId })
```
TeddyAndersson, line 30: What we could do here to reduce request time is to get all workflows, since the adapter does the same request either way but have a filter on the /workflow/{id} to only return the requested workflow. In other words both /workflows and /workflows/{id} sends the same request to viva from vada

TeddyAndersson: Might be a broader thing to discuss.

TeddyAndersson: Might be a cause for improved requests towards viva in in the viva adapter

services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
```diff
@@ -31,54 +53,35 @@ async function getAllUserCases(PK) {
...
-    }
-
-    if (!deepEqual(workflow.attributes, caseItem.details.workflow)) {
-      await syncWorkflowAndStatus(caseItem.PK, caseItem.SK, workflow.attributes);
-    }
-  }
+  return dynamoDb.call('query', params);
 }
 
 async function syncWorkflowAndStatus(PK, SK, workflow) {
```
TeddyAndersson, line 59: For me this function does many things, if we where to break out the different operations that is done here it would be more readable.

getStatusFromTypeCode
updateCaseWorkflow(PK, SK, workflow, status)

TeddyAndersson (approved):
Tested on local AWS environment and from my knowledge this works as intended.

PR #171, opened 2021-03-19
Feat/set initial case answers

jonatanhanson:
Added functionality to pre populate case answers with previous case answers and user information. 

## Explain your solution
Created a method that takes four arguments (initial forms, user object, previous case & form templates) and populates the initial forms with previous case data and user data. 
Which fields that will be pre populated can be defined from the formbuilder.

**Note** that the feature to load previous data is only activated to some of the fields(text, number, checkbox, repeaterField and editableList) in the formbuilder and will not work for rest of the fields. I'm not sure why we don't have it activated for all fields, could be a future task to fix. 

This new functionality is used by createCase + createVivaCase lambdas during the creation of new cases.

I have chosen to get previous case data from the latest updated case that also has status set to "closed". So this will not get previous answers from "ongoing" cases for example.

## How to test the changes?

First make sure you have a user, forms and a case(with status: closed) in your DB.
The form needs to have fields with "load previous" functionality activated. 

Deploy case and viva services. 
Make a POST request to create case endpoint.
Example body json:
`
{
  "statusType": "notStarted",
  "currentFormId": "SOME-FORM-ID",
  "provider": "VIVA",
  "details": {},
  "forms": {
    "SOME-FORM-ID {
      "answers": [],
      "currentPosition": {
        "index": 1,
        "level": 0,
        "currentMainStep": 0,
        "currentMainStepIndex": 0
      }
    }
  }
}
` 

If everything works the answers object should be populated with the data you have defined in form builder.  

## Anything else
I have only tested this with the createCase endpoint, because I dont know how to trigger event that creates a new VIVA case. So it would be great if someone could test that too.

### DIFF
--- a/services/cases-api/lambdas/createCase.js
+++ b/services/cases-api/lambdas/createCase.js
@@ -5,6 +5,7 @@ import { throwError } from '@helsingborg-stad/npm-api-error-handling';
 import uuid from 'uuid';
 
 import config from '../../../config';
+import * as dynamoDb from '../../../libs/dynamoDb';
 import * as response from '../../../libs/response';
 import { decodeToken } from '../../../libs/token';
 import { getItem, putItem } from '../../../libs/queries';
@@ -39,7 +40,15 @@ export async function main(event) {
   const PK = `USER#${personalNumber}`;
   const SK = `USER#${personalNumber}#CASE#${id}`;
 
-  const initialForms = populateFormAnswers(forms, personalNumber);
+  const userParams = {
+    TableName: config.users.tableName,
+    Key: {
+      personalNumber,
+    },
+  };
+
+  const user = await getUser(userParams);
+  const initialForms = populateFormAnswers(forms, user, personalNumber);
   console.log(util.inspect(initialForms, { showHidden: false, depth: null }));
 
   const timestampNow = Date.now();
@@ -109,4 +118,10 @@ async function parseJsonD(data) {
   } catch (error) {
     throwError(400, error.message);
   }
+}
+
+async function getUser(params) {
+  const [error, dbResponse] = await to(dynamoDb.call('get', params));
+  if (!dbResponse) throwError(error);
+  return dbResponse;
 }
--- a/services/cases-api/lambdas/createCase.js
+++ b/services/cases-api/lambdas/createCase.js
@@ -42,7 +42,8 @@ export async function main(event) {
 
   const user = await getUser(personalNumber);
   const formTemplates = await getFormTemplates(forms);
-  const initialForms = populateFormAnswers(forms, user, formTemplates);
+  const previousCase = await getLastUpdatedCase(personalNumber, provider);
+  const initialForms = populateFormAnswers(forms, user, formTemplates, previousCase);
   console.log(util.inspect(initialForms, { showHidden: false, depth: null }));
 
   const timestampNow = Date.now();
@@ -150,4 +151,32 @@ async function getFormTemplates(forms) {
   }
 
   return formTemplates;
+}
+
+async function getLastUpdatedCase(personalNumber, provider) {
+  const params = {
+    TableName: config.cases.tableName,
+    KeyConditionExpression: 'PK = :pk',
+    FilterExpression: 'begins_with(#status.#type, :statusTypeClosed) and #provider = :provider',
+    ExpressionAttributeNames: {
+      '#status': 'status',
+      '#type': 'type',
+      '#provider': 'provider',
+    },
+    ExpressionAttributeValues: {
+      ':pk': `USER#${personalNumber}`,
+      ':statusTypeClosed': 'closed',
+      ':provider': provider,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDb.call('query', params));
+  if (!dbResponse) {
+    console.error('(cases-api) DynamoDb query on cases table failed', error);
+    return;
+  }
+
+  const sortedCases = dbResponse.Items.sort((a, b) => b.updatedAt - a.updatedAt);
+
+  return sortedCases?.[0] || undefined;
 }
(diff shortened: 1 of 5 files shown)


zilaei:
Tested with the app and can verify that previews data is loaded in current case (for dynamic inputs).

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -152,3 +158,70 @@ async function putRecurringVivaCase(PK, workflowId, period) {
...
+    Key: {
+      personalNumber,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDB.call('get', params));
+  if (!dbResponse) {
+    console.error('(cases-api) DynamoDb query on users table failed', error);
+    return;
+  }
```
TeddyAndersson, line 175: I would rather see that we throw this error than return nothing at all. the console.error can be used in an catch further upstream.

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -152,3 +158,70 @@ async function putRecurringVivaCase(PK, workflowId, period) {
...
+      ':statusTypeClosed': 'closed',
+      ':provider': provider,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDB.call('query', params));
+  if (!dbResponse) {
+    console.error('(cases-api) DynamoDb query on cases table failed', error);
+    return;
+  }
```
TeddyAndersson, line 222: Throw error instead of return.

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -126,6 +127,11 @@ async function putRecurringVivaCase(PK, workflowId, period) {
     [completionFormId]: initialFormAttributes,
   };
 
+  const user = await getUser(PK);
+  const formTemplates = await getFormTemplates(initialForms);
+  const previousCase = await getLastUpdatedCase(PK, CASE_PROVIDER_VIVA);
```
TeddyAndersson, line 132: use await to syntax

TeddyAndersson: and logg the errors here in the main function

TeddyAndersson (changes requested):
The code works from the test that i made on my personal environment. Added some comments for code readability and consistency

services/cases-api/lambdas/createCase.js
```diff
@@ -104,3 +124,67 @@ async function parseJsonD(data) {
...
+    TableName: config.users.tableName,
+    Key: {
+      personalNumber,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDb.call('get', params));
+  if (!dbResponse) {
+    throwError(error.statusCode, error.message);
+  }
```
dannil76, line 139: To be consistent with how other lambdas is handling error we can write this like
`if (error) {
.....
}`

:)

services/cases-api/lambdas/createCase.js
```diff
@@ -104,3 +124,67 @@ async function parseJsonD(data) {
...
+      Key: {
+        PK: `FORM#${key}`,
+      },
+    };
+    const [error, dbResponse] = await to(dynamoDb.call('get', params));
+
+    if (!dbResponse) {
+      console.error('(cases-api) DynamoDb query on forms table failed', error);
+      continue;
+    }
```
dannil76, line 158: same comment as previous comment :)

services/cases-api/lambdas/createCase.js
```diff
@@ -104,3 +124,67 @@ async function parseJsonD(data) {
...
+      ':pk': PK,
+      ':statusTypeClosed': 'closed',
+      ':provider': provider,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDb.call('query', params));
+  if (!dbResponse) {
+    throwError(error.statusCode, error.message);
+  }
```
dannil76, line 185: same as previous comments :)

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -152,3 +172,68 @@ async function putRecurringVivaCase(PK, workflowId, period) {
...
+    TableName: config.users.tableName,
+    Key: {
+      personalNumber,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDB.call('get', params));
+  if (!dbResponse) {
+    throwError(error.statusCode, error.message);
+  }
```
dannil76, line 188: same feedback as previous comments on how error handling is done in other lambdas
check IF error is a little more telling :)

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -152,3 +172,68 @@ async function putRecurringVivaCase(PK, workflowId, period) {
...
+      Key: {
+        PK: `FORM#${key}`,
+      },
+    };
+    const [error, dbResponse] = await to(dynamoDB.call('get', params));
+
+    if (!dbResponse) {
+      console.error('(cases-api) DynamoDb query on forms table failed', error);
+      continue;
+    }
```
dannil76, line 207: same feedback as previous

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -152,3 +172,68 @@ async function putRecurringVivaCase(PK, workflowId, period) {
...
+      ':pk': PK,
+      ':statusTypeClosed': 'closed',
+      ':provider': provider,
+    },
+  };
+
+  const [error, dbResponse] = await to(dynamoDB.call('query', params));
+  if (!dbResponse) {
+    throwError(error.statusCode, error.message);
+  }
```
dannil76, line 234: same feedback as previous

dannil76 (changes requested):
some minor aesthetically request :)

services/cases-api/lambdas/createCase.js
jonatanhanson: Makes sense, fixed now

dannil76 (approved):
Brilliant :)

PR #172, opened 2021-03-24
fix: 🐛 PDF-ms npm packages version problems

dannil76:
What?
Removed npm packages (serverless and more) from local package.json because it's incompatible with newer version of serverless.yml format.

We should use serverless and other npm packages that is globaly installed at repo root.

Tested
Successfully test on personal aws sandbox

Howto test
Should work as before

### DIFF
--- a/services/pdf-ms/package.json
+++ b/services/pdf-ms/package.json
@@ -5,23 +5,14 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-g71nav Bugg: Fil rättigheter för PDF S3/Lambda](https://app.clickup.com/t/g71nav)

jonatanhanson (approved):
Works for me!

PR #178, opened 2021-04-13
Fix/viva ms sync workflow

dannil76:
For more background, see Clickup

For more background, see Clickup

## Explain your solution
For more background, see Clickup

## How to test the changes?
Checkout PR
Deploy viva-ms to your aws sandbox
For more background, see Clickup

### DIFF
--- a/services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
+++ b/services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
@@ -143,7 +143,9 @@ function makeArray(value) {
 
   if (Array.isArray(value)) {
     list = [...value];
-  } else if (value !== undefined) {
+  }
+
+  if (value !== undefined) {
     list.push(value);
   }
 
--- a/services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
+++ b/services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
@@ -10,11 +10,11 @@ import vivaAdapter from '../helpers/vivaAdapterRequestClient';
 export async function main(event) {
   const personalNumber = event.detail.user.personalNumber;
 
-  const [getUserCasesError, userCases] = await to(
-    getSubmittedAndOrProcessingUserCases(personalNumber)
+  const [getCasesError, userCases] = await to(
+    getCasesWithStatusSumbittedOrProcessing(personalNumber)
   );
-  if (getUserCasesError) {
-    throw ('(Viva-ms) DynamoDB query failed', getUserCasesError);
+  if (getCasesError) {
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-gf0ekn Bugg: Viva-MS sync workflow - Decision List ska alltid vara en lista](https://app.clickup.com/t/gf0ekn)

dannil76:
Nobody dare to test?

Sven65:
Tested according to instructions, did `sls remove` first to clean.

The wanted result wasn't achieved. No workflow gets added in DynamoDB.

services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
```diff
@@ -7,99 +7,147 @@ import * as dynamoDb from '../../../libs/dynamoDb';
...
-  const PK = `USER#${personalNumber}`;
 
-  const [getAllUserCasesError, allUserCases] = await to(getAllUserCases(PK));
-  if (getAllUserCasesError) {
-    return console.error('(Viva-ms) DynamoDB query failed', getAllUserCasesError);
+  const [getCasesError, userCases] = await to(
+    getCasesWithStatusSumbittedOrProcessing(personalNumber)
+  );
+  if (getCasesError) {
+    throw ('(Viva-ms) getCasesWithStatusSumbittedOrProcessing', getCasesError);
```
Sven65, line 17: This will not work as intended, as it'll just throw the value of `getCasesError`.
The `(Viva-ms) getCasesWithStatusSumbittedOrProcessing` part will **not** follow along in the error message.

For reference, try this in your browser console;

```js
const myError = "This is an error string"

try {
	throw ('I want this to log', myError)
} catch (e) {
	console.error(e)
}
```

All you'll get is the string `This is an error string`

Sven65 (changes requested):

services/viva-ms/lambdas/syncWorkflowAndSetStatus.js
TeddyAndersson: instead of throwing this error you could do something in the lines of:
```
if (getCasesError) {
   console.error('(Viva-ms) getCasesWithStatusSumbittedOrProcessing', error)
   return false 
}

TeddyAndersson, line 17: This throw syntax won't work as expected @Sven65 earlier mentioned

TeddyAndersson (changes requested):
This syntax was never fixed

TeddyAndersson (approved):

zilaei (approved):
👍

PR #181, opened 2021-04-22
fix(submitcompletion): setting attachment name to uploadedFilename

TeddyAndersson:
The filename attribute that is passed from the client to an item in DynamoDB Cases Table, was not
always set due to limit in the image library that is used. Since the viva adapter requires a name to
be passed on an attachment it would break. To fix this issue we now use the name of the uploaded
file in s3 instead, since that value will never be null


helsingborgsstad:
Task linked: [CU-hf1knt Inskick av ny komplettering till adaptern svarar med status 500 (Internal Server Error)](https://app.clickup.com/t/hf1knt)

Sven65 (approved):

PR #183, opened 2021-04-27
fix: more permissions for attachments

yo-l1982:
Fix missing permissions on attachments upload


KenthLjungHBG (approved):
Tested with own sandbox - fixed s3 upload 403 bug.

PR #185, opened 2021-05-07
feat(api): add API Key requirement to endpoints

Sven65:
More info at #CU-ht607d

### DIFF
--- a/services/bankid-api/serverless.yml
+++ b/services/bankid-api/serverless.yml
@@ -68,6 +68,7 @@ functions:
       - http:
           path: auth/bankid/auth
           method: post
+          private: true
           cors: true
   sign:
     handler: lambdas/sign.main
@@ -75,6 +76,7 @@ functions:
       - http:
           path: auth/bankid/sign
           method: post
+          private: true
           cors: true
   collect:
     handler: lambdas/collect.main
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-ht607d Slå på api-key i serverless.yml på samtliga endpoints](https://app.clickup.com/t/ht607d)

TeddyAndersson (commented):
Seems to work as expected, have you tested this with the app aswell?

services/bankid-api/serverless.yml
```diff
@@ -68,25 +68,29 @@ functions:
...
           path: auth/bankid/collect
           method: post
+          private: true
           cors: true
   cancel:
     handler: lambdas/cancel.main
     events:
       - http:
           path: auth/bankid/cancel
           method: post
```
dannil76, line 94: Is there a reason not to have the parameter authorizer defined?
See cases-api for reference.

Sven65: I'm unsure whether or not it was needed. Seeing as the task only was for API Keys, and that we need to use the bankID to login, I opted to not include it.

Sven65:
> Seems to work as expected, have you tested this with the app aswell?

I haven't tested it fully yet, but from the testing I have done, it works as intended.

PR #189, opened 2021-05-24
Fix/viva ms decision

dannil76:
## Feature description
Read more in clickup

## Solution description
Extracted viva-ms lambda syncWorkflowAndUpdateStatus in two separate lambdas decideCaseStatus resp. syncWorkflow

syncWorkflow - update case attribute: workflow on cases where status is submitted or processing. For each update the lambda emits an event vivaMsSyncWorkflowSuccess

decideCaseStatus - listening on event vivaMsSyncWorkflowSuccess and update case status based on new workflow data

Lambdas in viva-ms decideCaseStatus and syncWorkflow

No

No

## Are your code strutured in a way so that reviewers can understand it?
I hope

- [x] Your local Viva adapter environment.


helsingborgsstad:
Task linked: [CU-jb97f3 Status för beslut tillsammans med utbetalning ska visas när utbetalning är gjord i Viva](https://app.clickup.com/t/jb97f3)

services/viva-ms/lambdas/decideCaseStatus.js
```diff
@@ -0,0 +1,85 @@
...
+  const decisionList = makeArray(workflowAttributes.decision?.decisions?.decision);
+  const paymentList = makeArray(workflowAttributes.payments?.payment);
+  const calculation = workflowAttributes.calculations?.calculation;
+
+  let decisionStatus = 0;
+  let newStatusType = '';
+
+  if (decisionList != undefined && decisionList.length > 0) {
+    decisionList.forEach(decision => {
+      decisionStatus = decisionStatus | parseInt(decision.typecode, 10);
```
KenthLjungHBG, line 52: It's not clear how`decisionStatus`/`typecode` works. I assume they're some kind of flags? A suggestion would be to use named constants for these magic values.

dannil76: It's tricky yes. :)
The typeCode can be one of 01, 02 or 03.
By using the bitewise OR operator we can figure out the correct statuscode.

I'll be happy to explain in more detail IRL. :)

KenthLjungHBG (approved):
Based on code it looks good, but I haven't been able to test.

PR #190, opened 2021-05-26
Feat/case initial expire time

TeddyAndersson:
* Added initial expirationTime when a viva case is created.
* Fixed inconsistency in retrieval of personalNumber from PK.

These changes have been made to remove open cases that have not been started currently open for a users from being able to start an old case.

## Explain your solution
Calculated the expirationTime from the current point in time when a case is created to get a future timestamp for when the case should be deleted. This timestamp is set on the case when adding it to dynamodb, dynamodb then handles the deletion when the expiration time is surpassed.

## How to test the changes?
1. Checkout this branch.
2. Navigate to /services/viva-ms `yarn install`.
3. Deploy the microservice `sls deploy`.
4. Trigger the createVivaCase lambda by an eventbridge event to create a new case.


KenthLjungHBG (commented):
Verified working in my sandbox.

PR #193, opened 2021-06-04
Perf/update case status

TeddyAndersson:

Removed support for setting the status on a case from request attributes, the status is now instead dynamically set based on attributes in the request data.

Added new status for case active:signature:completed
Checking if the case exists before trying to preform updates.
Added support for conditional retrieving of the case status based on answers and signatures attributes in the json payload.
Updating the case status if there is any new status to be retrieved.

## Explain your solution

Earlier we allowed the setting of a case status by passing a value on the attribute statusType
in the json payload, this behavior is now deprecated. 

Now instead the lambda looks on the data in the json payload to figure out if the case status should be updated and what that status is. 

The lambda can currently set three statuses on a case active:ongoing, active:signature:completed,
active:submitted based on the answers attribute and a new attribute called signature from the json
payload.

Statuses are set when a certain condition is true:
active:ongoing is set when answers in the payload are encrypted.
active:signature:completed is set when the answers in the payload are encrypted and a the signature is completed 
active:submitted is set when a answers are decrypted and the signature is completed.

## How to test the changes?

Concrete example:
1. Checkout this branch.
2. Recommended to use your own AWS sandbox environment.
3. Deploy the cases service.
4. Add/Create an initial case in your DynamoDB
5. Send put request towards /cases/{id} (id should be the id from your new case)

ongoing status
```JSON 
{
   answers: { encryptedAnswers: "an_encrypted_string"}
   signature: { successfull: false}
   ...
}
```

signature completed status
```JSON 
{
   answers: { encryptedAnswers: "an_encrypted_string"}
   signature: { successfull: true}
   ...
}
```

submitted status
```JSON 
{
   answers: [{...}, {...}]
   signature: { successfull: true}
   ...
}
```

- [X] AWS Sandbox environment.

## Anything else
BREAKING CHANGE: All clients that uses the /cases with the PUT method will need to update their
requests in order to comply with these new changes.

### DIFF
--- a/libs/caseStatuses.js
+++ b/libs/caseStatuses.js
@@ -11,9 +11,9 @@ const statuses = [
       'Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade.',
   },
   {
-    type: 'active:signed:viva',
+    type: 'active:signature:completed',
     name: 'Signerad',
-    description: 'Ansökan är signerad, ladda upp filer & dokument för att skicka in ansökan.',
+    description: 'Ansökan är signerad',
   },
   {
     type: 'active:submitted',
--- a/libs/caseStatuses.js
+++ b/libs/caseStatuses.js
@@ -11,7 +11,7 @@ const statuses = [
       'Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade.',
   },
   {
-    type: 'active:signature:completed',
+    type: 'active:signed',
     name: 'Signerad',
     description: 'Ansökan är signerad',
   },
--- a/services/cases-api/helpers/dynamo.js
+++ b/services/cases-api/helpers/dynamo.js
@@ -1,0 +1,19 @@
+import dynamoDb from '../../../libs';
+import config from '../../../config';
+
+export async function getCaseWhereUserIsApplicant(caseId, userPersonalNumber) {
+  const PK = `USER#${userPersonalNumber}`;
+  const SK = `${PK}#CASE#${caseId}`;
+
+  const dynamoDbQueryParams = {
+    TableName: config.cases.tableName,
+    ExpressionAttributeValues: {
+      ':pk': PK,
+      ':sk': SK,
+    },
+    KeyConditionExpression: 'PK = :pk AND SK = :sk',
+    Limit: 1,
+  };
+
+  return dynamoDb.call('query', dynamoDbQueryParams);
+}
--- a/services/cases-api/lambdas/updateCase.js
+++ b/services/cases-api/lambdas/updateCase.js
@@ -179,19 +179,19 @@ function getNewCaseStatus(answers, signature) {
   return getStatusByType(statusType);
 }
 
-function isAnswersEncrypted(answers) {
+function areAnswersEncrypted(answers) {
   const keys = Object.keys(answers);
   return keys.length === 1 && keys.includes('encryptedAnswers');
 }
 
 function isOngoing(answers) {
-  return answers && isAnswersEncrypted(answers);
+  return answers && areAnswersEncrypted(answers);
 }
 
 function isSignatureCompleted(answers, signature) {
-  return signature.successfull && isAnswersEncrypted(answers);
+  return signature.success && areAnswersEncrypted(answers);
 }
 
 function isSubmitted(answers, signature) {
-  return signature.successfull && !isAnswersEncrypted(answers);
+  return signature.success && !areAnswersEncrypted(answers);
 }
--- a/services/cases-api/lambdas/updateCase.js
+++ b/services/cases-api/lambdas/updateCase.js
@@ -1,5 +1,9 @@
 import to from 'await-to-js';
-import { throwError, BadRequestError } from '@helsingborg-stad/npm-api-error-handling';
+import {
+  throwError,
+  BadRequestError,
+  NotFoundError,
+} from '@helsingborg-stad/npm-api-error-handling';
 
 import config from '../../../config';
 import { putEvent } from '../../../libs/awsEventBridge';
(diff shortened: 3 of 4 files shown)


zilaei:
Is package @helsingborg-stad/npm-api-error-handling updated with latest code? Can't resolve 'NotFoundError'.

services/cases-api/lambdas/updateCase.js
```diff
@@ -8,18 +12,28 @@ import * as dynamoDb from '../../../libs/dynamoDb';
...
+  const [getCaseError] = await to(dynamo.getCaseWhereUserIsApplicant(id, personalNumber));
+  if (getCaseError) {
+    return response.failure(
+      new NotFoundError('The user does not have any case with the provided case id')
+    );
   }
 
-  const { provider, statusType, details, currentFormId, currentPosition, answers } =
-    requestJsonBody;
+  const newCaseStatus = getNewCaseStatus(answers, signature);
```
KenthLjungHBG, line 34: This uses the `answers` from the request body, meaning that even if this update call is only to update signatures, the caller has to have and send the (potentially encrypted) answers as well. The rest of the function implies that `answers` is optional. 

In that case (and in general to avoid unintended exploits) we should null-check `answers` and `signature` before running `getNewCaseStatus`, otherwise there exists scenarios where for example `answers` is null, and `signature` is successful, causing the status to be `active:submitted`.

An alternative solution (if the intention is ever to send `signature` without `answers`) is to fallback to the existing `answers` in the case when calling `getNewCaseStatus`.

KenthLjungHBG (changes requested):

services/cases-api/helpers/dynamo.js
```diff
@@ -0,0 +1,19 @@
...
+    ExpressionAttributeValues: {
+      ':pk': PK,
+      ':sk': SK,
+    },
+    KeyConditionExpression: 'PK = :pk AND SK = :sk',
+    Limit: 1,
+  };
+
+  return dynamoDb.call('query', dynamoDbQueryParams);
+}
```
dannil76, line 19: Will this support the new GSI implementation on cases db?

dannil76 (changes requested):

TeddyAndersson:
> Is package @helsingborg-stad/npm-api-error-handling updated with latest code? Can't resolve 'NotFoundError'.

It should be fine, could you paste the error output from your terminal? :)

services/cases-api/helpers/dynamo.js
TeddyAndersson: There should not be any interference with the new GSI implementation

services/cases-api/lambdas/updateCase.js
```diff
@@ -8,18 +12,28 @@ import * as dynamoDb from '../../../libs/dynamoDb';
...
   if (!id) {
-    return response.failure(new BadRequestError('missing [id] in query path'));
+    return response.failure(new BadRequestError('Missing [id] in path /cases/{id}'));
+  }
+
+  const [getCaseError] = await to(dynamo.getCaseWhereUserIsApplicant(id, personalNumber));
+  if (getCaseError) {
+    return response.failure(
+      new NotFoundError('The user does not have any case with the provided case id')
+    );
```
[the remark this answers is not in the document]
TeddyAndersson: Added this DB call with the notion that we will need to verify the signature in a near future towards the case in the database. So totally get your point :)

Can withdraw these changes, if we feel it to be unnecessary at this point :)

libs/caseStatuses.js
```diff
@@ -11,9 +11,9 @@ const statuses = [
       'Du har påbörjat en ansökan. Du kan öppna din ansökan och fortsätta där du slutade.',
   },
   {
-    type: 'active:signed:viva',
+    type: 'active:signature:completed',
```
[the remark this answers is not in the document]
TeddyAndersson: I guess it somehow follows the status in terms of hierarchy of information. But i can see the confusion. My thought was to categories the signature status into a group that then have two states completed (active:signature:completed) and pending (active:signature:pending).

Mostly because it felt weird to add a status with camelcase like active:pendingSignature. But i'm open for suggestion :)

services/cases-api/lambdas/updateCase.js
TeddyAndersson: Yeah, this entire lambda is a bit strange. Since the main usage is to update form answers within a case, don't know if we do run updates towards a case in any other way. So we might want to require answers to be submitted in the payload?

If this is the case then it would make more sense to have an /cases/{id}/forms/{id} endpoint instead?

TeddyAndersson: But i agree the status behaviour that i have added in this function should have some sort of data validation :)

services/cases-api/lambdas/updateCase.js
```diff
@@ -8,18 +12,28 @@ import * as dynamoDb from '../../../libs/dynamoDb';
...
 
 export async function main(event) {
   const requestJsonBody = JSON.parse(event.body);
   const { id } = event.pathParameters;
 
+  const { provider, details, currentFormId, currentPosition, answers, signature } = requestJsonBody;
+
   if (!id) {
-    return response.failure(new BadRequestError('missing [id] in query path'));
+    return response.failure(new BadRequestError('Missing [id] in path /cases/{id}'));
```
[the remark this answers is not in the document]
TeddyAndersson: since the lambda is connected and expects an http event and path parameters maybe `Missing required path parameter "id"` would be more clear?

libs/caseStatuses.js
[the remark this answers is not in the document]
KenthLjungHBG: Agree that camel case feels a bit weird in this context, however we already have several camel case statuses such as `partiallyApproved`, `completionRequired` etc. so for consistency I prefer that approach.

Grouping by `signature` also seems like a good idea however, so perhaps this design discussion needs more input.

services/cases-api/lambdas/updateCase.js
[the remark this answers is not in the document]
KenthLjungHBG: In that case I think it's fine to keep as is for now 🙂.

services/cases-api/lambdas/updateCase.js
KenthLjungHBG: Separating concerns is a good idea! Especially as the project grows. But for now I think adding some simple data validation is the way to go, without over-engineering too much 🙂.

TeddyAndersson: Ill introduce the schema validation that we have on other endpoints to this endpoint in order to have a consistent data validation (y)

libs/caseStatuses.js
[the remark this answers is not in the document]
TeddyAndersson: I can change it to active:signed, for these changes then we can tackle this issue when we implement the pending/completed flow for signatures

services/cases-api/lambdas/updateCase.js
```diff
@@ -12,13 +12,33 @@ import * as dynamoDb from '../../../libs/dynamoDb';
...
+    provider,
+    details,
+    currentFormId,
+    currentPosition,
+    answers,
+    signature,
+  } = validatedEventBody;
 
   if (!id) {
     return response.failure(new BadRequestError('Missing [id] in path /cases/{id}'));
```
dannil76: Should this be here? You check if id at row 23.

services/cases-api/lambdas/updateCase.js
```diff
@@ -179,19 +199,24 @@ function getNewCaseStatus(answers, signature) {
   return getStatusByType(statusType);
 }
 
-function isAnswersEncrypted(answers) {
+function areAnswersEncrypted(answers) {
```
dannil76, line 202: like is over are :)
isEncrypted() is maybe enought because you put answers in args?

KenthLjungHBG (approved):

PR #196, opened 2021-06-09
Feat/get person data from viva

TeddyAndersson:
## Feature description
Adding support for getting viva person information (case, application) from the helpers library.

## Solution description
A new helper function `getPerson` have been added in `vivaAdapterRequestClient.js`. The function gets all information on a person and formats the response for easier management, since the response data contains meny nested objects before you can get the data you want.

How to use the new request function
```JS
import vivaAdapter from 'vivaAdapterRequestClient'

async function simpleExample() {
  const personalNumber = 123456789
  const requestResponse = await vivaAdapter.person.get(personalNumber)
  // ...do something with the the requestResponse
}
```

viva-ms/helpers/vivaAdapterRequestClient
viva-ms/lambdas/createVivaCase

## Is there any existing behavior change of other features due to this code change?
Yes, The createVivaCase have been refactored to use the person.get function instead of application.get, since we will need both the application data and the case data for a person.

## How to test
1. Checkout this branch
2. Deploy the viva-ms service.
3. Fire an event to trigger the createVivaCase lambda. (if there are an open period this code will run)
4. If a new case is created in your database then everything is fine.

- [x] Your personal AWS environment

### DIFF
--- a/services/viva-ms/lambdas/createVivaCase.js
+++ b/services/viva-ms/lambdas/createVivaCase.js
@@ -46,9 +46,9 @@ export async function main(event) {
     );
   }
 
-  const [getApplicationError, vivaPerson] = await to(vivaAdapter.person.get(user.personalNumber));
-  if (getApplicationError) {
-    return console.error('(Viva-ms) Viva Get Application Request', getApplicationError);
+  const [getPersonError, vivaPerson] = await to(vivaAdapter.person.get(user.personalNumber));
+  if (getPersonError) {
+    return console.error('(Viva-ms) Viva Get Application Request', getPersonError);
   }
 
   if (!vivaPerson.application || !vivaPerson.application.period) {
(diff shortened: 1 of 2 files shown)


services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -46,26 +46,24 @@ export async function main(event) {
     );
   }
 
-  const [getApplicationError, application] = await to(
-    vivaAdapter.application.get(user.personalNumber)
-  );
+  const [getApplicationError, vivaPerson] = await to(vivaAdapter.person.get(user.personalNumber));
```
KenthLjungHBG, line 49: Rename `getApplicationError` to `getPersonError`?

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -46,26 +46,24 @@ export async function main(event) {
...
-  if (!application || !application.period) {
+  if (!vivaPerson.application || !vivaPerson.application.period) {
     return console.error('(Viva-ms) Viva Application Period not present in response, aborting');
   }
 
-  if (!application || !application.workflowid) {
+  if (!vivaPerson.application || !vivaPerson.application.workflowid) {
     return console.error(
-      `(Viva-ms) Viva Application WorkflowId ${application.workflowid} not present in response, aborting`
+      `(Viva-ms) Viva Application WorkflowId ${vivaPerson.application.workflowid} not present in response, aborting`
```
KenthLjungHBG, line 60: Add `?` to `application` for the error message, otherwise this will fail in case it's invalid.

KenthLjungHBG (changes requested):

KenthLjungHBG (approved):

dannil76 (approved):
Love you!

PR #197, opened 2021-06-09
Feat/viva ms set keys pk gsi

dannil76:
See clickup link below

### DIFF
--- a/services/viva-ms/lambdas/createVivaCase.js
+++ b/services/viva-ms/lambdas/createVivaCase.js
@@ -82,13 +82,16 @@ export async function main(event) {
 }
 
 async function getUserCaseFilteredOnWorkflowId(vivaPerson) {
+  const personalNumber = String(vivaPerson.case.client.pnumber).replace(/\D/g, '');
+  const workflowId = vivaPerson.application.workflowid;
+
   const params = {
     TableName: config.cases.tableName,
     KeyConditionExpression: 'PK = :pk',
     FilterExpression: 'details.workflowId = :workflowId',
     ExpressionAttributeValues: {
-      ':pk': `USER#${vivaPerson.case.client.pnumber}`,
-      ':workflowId': vivaPerson.application.workflowid,
+      ':pk': `USER#${personalNumber}`,
+      ':workflowId': workflowId,
     },
     Limit: 1,
   };
--- a/services/viva-ms/lambdas/createVivaCase.js
+++ b/services/viva-ms/lambdas/createVivaCase.js
@@ -173,12 +173,60 @@ async function putRecurringVivaCase(vivaPerson) {
     },
   };
 
+  const casePersonList = getCasePersonList(vivaPerson);
+  caseItemPutParams['persons'] = casePersonList;
+
+  const casePersonCoApplicant = getUserByRole(casePersonList, 'coApplicant');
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-kq7b10 Spara case keys(PK resp. GSI1) från viva-case-attributen: client.pnumber och persons.person.pnumber](https://app.clickup.com/t/kq7b10)

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -169,14 +177,64 @@ async function putRecurringVivaCase(PK, workflowId, period) {
...
+    client: 'applicant',
+    partner: 'coApplicant',
+    child: 'children',
+  };
+
+  const casePersonList = vivaPersonList.map(person => {
+    const { pnumber, fname: firstName, lname: lastName, type } = person;
+    const personalNumber = String(pnumber).replace(/\D/g, '');
+
+    const role = Object.keys(roleTranslateList).includes(type) && roleTranslateList[type];
```
KenthLjungHBG, line 225: If type is invalid then role becomes a bool (`false`), and `<bool> && <string|undefined>` is not very clear, given javascript's equality inconsistency.

Suggestion: change to `const role = roleTranslateList[type] || ""`.

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -59,58 +59,65 @@ export async function main(event) {
...
 
-  const [dynamoQueryError, queryCasesResult] = await to(dynamoDB.call('query', params));
-  if (dynamoQueryError) {
-    throw dynamoQueryError;
+  const [queryCasesError, queryCasesResult] = await to(dynamoDB.call('query', params));
+  if (queryCasesError) {
+    throw queryCasesError;
   }
 
   return queryCasesResult.Items;
```
KenthLjungHBG, line 104: Since we're only fetching one case we should not return a list. 

Suggestion: change to `return queryCasesResult.Items[0] || null`.

helsingborgsstad:
Task linked: [CU-kq7vf3 Skapa nytt attribut på case - persons. Spara vilka som ska signera](https://app.clickup.com/t/kq7vf3)

KenthLjungHBG (changes requested):
See comments.

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -59,7 +58,7 @@ export async function main(event) {
     return console.error('(Viva-ms) Viva Application WorkflowId not present in response, aborting');
   }
 
-  const [getUserCaseFilteredOnWorkflowIdError, caseItem] = await to(
+  const [getUserCaseFilteredOnWorkflowIdError, caseItemExists] = await to(
```
KenthLjungHBG, line 61: Naming it `Exists` implies that it is a bool. `caseItem` is fine.

KenthLjungHBG (changes requested):

services/viva-ms/lambdas/createVivaCase.js
dannil76: You're a true champ. Tx! Fix this now

KenthLjungHBG (approved):

PR #201, opened 2021-06-14
Feat/update cases status

TeddyAndersson:

Removed support for setting the status on a case from request attributes, the status is now instead dynamically set based on attributes in the request data.

Added new status for case active:signature:completed
Checking if the case exists before trying to preform updates.
Added support for conditional retrieving of the case status based on answers and signatures attributes in the json payload.
Updating the case status if there is any new status to be retrieved.

## Explain your solution

Earlier we allowed the setting of a case status by passing a value on the attribute statusType
in the json payload, this behavior is now deprecated. 

Now instead the lambda looks on the data in the json payload to figure out if the case status should be updated and what that status is. 

The lambda can currently set three statuses on a case active:ongoing, active:signature:completed,
active:submitted based on the answers attribute and a new attribute called signature from the json
payload.

Statuses are set when a certain condition is true:
active:ongoing is set when answers in the payload are encrypted.
active:signature:completed is set when the answers in the payload are encrypted and a the signature is completed 
active:submitted is set when a answers are decrypted and the signature is completed.

## How to test the changes?

Concrete example:
1. Checkout this branch.
2. Recommended to use your own AWS sandbox environment.
3. Deploy the cases service.
4. Add/Create an initial case in your DynamoDB
5. Send put request towards /cases/{id} (id should be the id from your new case)

ongoing status
```JSON 
{
  "currentFormId": REPLACE_WITH_A_CURRENT_FORM_ID,
   "currentPosition": {
		"index": 0,
		"level": 0,
		"currentMainStep": 0,
		"currentMainStepIndex": 0
   },
   "answers": { "encryptedAnswers": "an_encrypted_string"},
   "signature": { "success": false}
}
```

signature completed status
```JSON 
{
  "currentFormId": REPLACE_WITH_A_CURRENT_FORM_ID,
   "currentPosition": {
		"index": 0,
		"level": 0,
		"currentMainStep": 0,
		"currentMainStepIndex": 0
   },
   "answers": { "encryptedAnswers": "an_encrypted_string"},
   "signature": { "success": true}
}
```

submitted status
```JSON 
{
   "currentFormId": REPLACE_WITH_A_CURRENT_FORM_ID,
   "currentPosition": {
		"index": 0,
		"level": 0,
		"currentMainStep": 0,
		"currentMainStepIndex": 0
   },
   "answers": [{...}, {...}],
   "signature": { "success": true}
}
```

- [X] AWS Sandbox environment.

## Anything else
BREAKING CHANGE: All clients that uses the /cases/{id} with the PUT method will need to update their
requests in order to comply with these new changes.

### DIFF
--- a/services/cases-api/helpers/schema.js
+++ b/services/cases-api/helpers/schema.js
@@ -7,7 +7,7 @@ const uuid = Joi.string().guid({
 
 const caseProvider = Joi.string().valid(CASE_PROVIDER_VIVA);
 
-const caseAnswers = Joi.array().items(
+const answers = Joi.array().items(
   Joi.object({
     field: Joi.object({
       id: Joi.string().required(),
@@ -17,22 +17,39 @@ const caseAnswers = Joi.array().items(
   })
 );
 
+const encryptedAnswers = Joi.object({
+  encryptedAnswers: Joi.string(),
+});
+
+const formCurrentPosition = Joi.object({
+  index: Joi.number().required(),
+  level: Joi.number().required(),
+  currentMainStep: Joi.number().required(),
+  currentMainStepIndex: Joi.number().required(),
+});
+
+const signature = Joi.object({
+  success: Joi.bool().required(),
+});
+
 const form = Joi.object({
-  answers: caseAnswers.allow(),
-  currentPosition: Joi.object({
-    index: Joi.number().required(),
-    level: Joi.number().required(),
-    currentMainStep: Joi.number().required(),
-    currentMainStepIndex: Joi.number().required(),
-  }).required(),
+  answers: answers.allow(),
+  currentPosition: formCurrentPosition.required(),
 });
 
 const caseValidationSchema = Joi.object({
-  statusType: Joi.string().valid('notStarted').required(),
+  statusType: Joi.string().valid('notStarted'),
   currentFormId: uuid.required(),
   provider: caseProvider.required(),
   details: Joi.object().allow(),
   forms: Joi.object().pattern(/^/, [uuid.required(), form.required()]),
+});
+
+export const updateCaseValidationSchema = Joi.object({
+  currentFormId: uuid.required(),
+  answers: [answers.allow(), encryptedAnswers.allow()],
+  currentPosition: formCurrentPosition.required(),
+  signature: signature.optional(),
 });
 
 export default caseValidationSchema;
--- a/services/cases-api/lambdas/updateCase.js
+++ b/services/cases-api/lambdas/updateCase.js
@@ -1,5 +1,9 @@
 import to from 'await-to-js';
-import { throwError, BadRequestError } from '@helsingborg-stad/npm-api-error-handling';
+import {
+  throwError,
+  BadRequestError,
+  ResourceNotFoundError,
+} from '@helsingborg-stad/npm-api-error-handling';
 
 import config from '../../../config';
 import { putEvent } from '../../../libs/awsEventBridge';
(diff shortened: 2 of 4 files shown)


helsingborgsstad:
Task linked: [CU-kx7kg3 Hantera status förändringar på api nivå när ett ärende i aws uppdateras](https://app.clickup.com/t/kx7kg3)

services/cases-api/helpers/schema.js
```diff
@@ -17,22 +17,39 @@ const caseAnswers = Joi.array().items(
...
   provider: caseProvider.required(),
   details: Joi.object().allow(),
   forms: Joi.object().pattern(/^/, [uuid.required(), form.required()]),
 });
 
+export const updateCaseValidationSchema = Joi.object({
+  currentFormId: uuid.required(),
+  answers: [answers.allow(), encryptedAnswers.allow()],
+  currentPosition: formCurrentPosition.required(),
+  signature: signature.optional(),
```
KenthLjungHBG, line 52: Missing `provider` and `details`.

services/cases-api/lambdas/updateCase.js
```diff
@@ -9,17 +13,44 @@ import { decodeToken } from '../../../libs/token';
...
+  const { personalNumber } = decodeToken(event);
 
   if (!id) {
-    return response.failure(new BadRequestError('missing [id] in query path'));
+    return response.failure(new BadRequestError('Missing required path parameter "id"'));
+  }
+
+  const { error, validatedEventBody } = updateCaseValidationSchema.validate(requestJsonBody, {
+    abortEarly: false,
+  });
```
KenthLjungHBG, line 30: Incorrect use of `validate`. Should be:

```javascript
  const { error, value: validatedEventBody } = updateCaseValidationSchema.validate(
    requestJsonBody,
    {
      abortEarly: false,
    }
  );
```

Currently, `validatedEventBody` is always `undefined`.

services/cases-api/lambdas/updateCase.js
```diff
@@ -9,17 +13,44 @@ import { decodeToken } from '../../../libs/token';
...
+  if (error) {
+    response.failure(new BadRequestError(error.message.replace(/"/g, "'")));
```
KenthLjungHBG, line 32: Missing `return`, causing process to keep going regardless of error.

services/cases-api/lambdas/updateCase.js
```diff
@@ -9,17 +13,44 @@ import { decodeToken } from '../../../libs/token';
...
+  }
+
+  const {
+    provider,
+    details,
+    currentFormId,
+    currentPosition,
+    answers,
+    signature,
+  } = validatedEventBody;
```
KenthLjungHBG, line 42: I couldn't run `sls deploy` without running `prettier` on this part first, as it throws an error here. 

`prettier` should be always be used regardless anyway. Make sure to setup it to run on save for your IDE (VS Code has a prettier extension, then set `prettier` as your default formatter).

services/cases-api/lambdas/updateCase.js
```diff
@@ -9,17 +13,44 @@ import { decodeToken } from '../../../libs/token';
...
+  const [getCaseError] = await to(getCaseWhereUserIsApplicant(id, personalNumber));
+  if (getCaseError) {
+    return response.failure(
+      new ResourceNotFoundError('The user does not have any case with the provided case id')
+    );
   }
 
-  const { provider, statusType, details, currentFormId, currentPosition, answers } =
-    requestJsonBody;
+  const newCaseStatus = getNewCaseStatus(answers, signature);
```
KenthLjungHBG, line 51: `signature` is only used here; it is never added/updated to the case (I assume it should be, like the other properties). Add `if (signature) { UpdateExpression.push(...) }` etc. together with the rest further down.

KenthLjungHBG (changes requested):

services/cases-api/lambdas/updateCase.js
TeddyAndersson: Had problems with prettier, seems like it was a mismatch between how serverless bundle runs. Turning prettier of and letting eslint handle the formatting fixed this issue for me. But happy for other suggestions, since not having prettier enabled in vscode is not ideal :)

services/cases-api/lambdas/updateCase.js
TeddyAndersson: Signature in this first implementation do not require any saving towards the case-database. But @dannil76 is working on an improved signature for multiple persons, where the signature will be updated for each person in the persons attribute on a case (y)

services/cases-api/helpers/schema.js
TeddyAndersson: Good point, then my next question are if these attributes are passed in any request from the app/client? otherwise it might be better to remove the option to update these values :)

services/cases-api/lambdas/updateCase.js
KenthLjungHBG: My suggestion would be to never let ESLint handle formatting (only linting), and let Prettier handle all formatting, and then apply the same eslintconfig and prettierrc across all js/ts projects in the future. But for now, if it deploys then it's fine! 🙂 Others might have to verify (or perhaps we can disable/ignore prettier errors on deploy; not sure how it is setup).

services/cases-api/lambdas/updateCase.js
TeddyAndersson: Might be something else that makes this act weird, since the documentation of validate clearly states the return value of the function to be an object with the attributes error and value. 

https://joi.dev/api/?v=17.4.0

services/cases-api/helpers/schema.js
KenthLjungHBG: Good counter-point, another solution would then be to remove `provider` and `details` from `updateCase` 👍 .

KenthLjungHBG (approved):

PR #203, opened 2021-06-16
Feat/signature status handling

dannil76:
## Feature description
User case (dynamoDn cases table) includes a new attribute - persons. Persons is a collection of users that are related to the case. One or more persons can be required for signing the case before submitting it.

Persons concists of one main person whose role is as the owner of the case - applicant. One secondary person whose role is as a co-applicant in a case. There can also be other roles like children etc.

In this particular situation persons is related to how a Viva case is constructed. The applicant and the co-applicant person object contains a hasSigned attribute and the purpose of this is to make it possible for ie. a smart phone application to implement some sort of signing mechanism where the case is required to be signed by one or more applicants before it's ready to be submitted

## Solution description
Signature is updated in connection with information being updated on a case. This is done by sending an extra attribute 'signature' in the PUT request to the endpoint /cases/{id}.

When a signature is sent the underlying lambda function - updateCase checks the attribute persons that are on a case to validate if the case has all the signatures needed to allow the user to submit the case.

The cases-api service - updateCase lambda

New attribute 'persons' and 'signature' on cases dynamoDb table

The endpoint /cases/{id} request body is in someway altered. Required PUT request body is shown below.
```
{
  "currentFormId": "nnnn",
  "currentPosition": {
    ...
  },
  "signature": {
    "success": true|false
  },
  "answers": [...]
}
```

## How to test
Verify attribute GSI1 is set to USER#(COAPPLICANT PNR) on the appalicant case with PK = USER#(APPLICANT PNR)

Make sure case contains the persons attribute
```
persons: [
  {
    personalNumber: (applicant),
    firstName: some-name,
    lastName: some-name,
    hasSigned: false,
    role: applicant
  },
  {
    personalNumber: (coApplicant),
    firstName: some-name,
    lastName: some-name,
    hasSigned: false,
    role: coApplicant
  },
]
```

Case status type set to notStarted
```
{
  type: 'notStarted:viva',
  name: 'Öppen',
  description: 'Ansökan är öppen. Du kan nu söka ekonomiskt bistånd för perioden.',
}
```
Answers should be encrypted or paste in below example answers via dynamoDb console
```
answers: [
  {
    'encryptedAnswers': 'abc123',
  }
]
```

Update case with PUT request via endpoint /case/{id} with the example request body from above and set signature.success to true. Send request using user appalicant access token. Repeate this request using coappalicant access token
```
...
"signature": {
  "success": true
},
...
```
Check case status


helsingborgsstad:
Task linked: [CU-kq7jae Uppdatera status på ärendet baserat på vilka personer som har signerat inväntar signering](https://app.clickup.com/t/kq7jae)

services/cases-api/lambdas/updateCase.js
```diff
@@ -25,32 +27,44 @@ export async function main(event) {
...
-  // Create update expression
-  const UpdateExpression = ['SET updatedAt = :newUpdatedAt'];
+  const UpdateExpression = ['updatedAt = :newUpdatedAt'];
   const ExpressionAttributeNames = {};
   const ExpressionAttributeValues = { ':newUpdatedAt': Date.now() };
 
+  const updatedUserCasePeople = userCase.persons?.map(person => {
+    const newPerson = { ...person };
+    if (newPerson.personalNumber === personalNumber) {
+      newPerson.hasSigned = signature?.success || false;
```
KenthLjungHBG, line 55: In case `signature` is not provided, this would reset it to `false`. Is this intended?

KenthLjungHBG (changes requested):

services/cases-api/helpers/statusCheckCondition.js
```diff
@@ -0,0 +1,53 @@
...
+  return peopleWhoMustSign.every(person => person.hasSigned === false);
+}
+
+function hasSomeOneSigned(people) {
+  const peopleWhoMustSign = selectPeopleWhoMustSign(people);
+  return peopleWhoMustSign.some(person => person.hasSigned === true);
+}
+
+function selectPeopleWhoMustSign(people) {
+  return people.filter(person => person?.hasSigned);
```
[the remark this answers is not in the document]
TeddyAndersson: We could also check if the key exits in the object by doing:
`return people.filter(person => 'hasSigned' in person);`

services/cases-api/lambdas/updateCase.js
TeddyAndersson: Good point, do not know if this is behaviour is wanted. My thought is that if a user updates the answers on a certain case, we want to reset the signature since the answers is updated and therefore the signature is for a previous collection of answers.

services/cases-api/helpers/statusCheckCondition.js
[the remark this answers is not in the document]
TeddyAndersson: I guess the best way according to [this]( https://stackoverflow.com/questions/455338/how-do-i-check-if-an-object-has-a-key-in-javascript) stackoverflow thread to check if a key exists in an object is to use:

`Object.prototype.hasOwnProperty.call(person, 'hasSigned');`

services/cases-api/lambdas/updateCase.js
```diff
@@ -88,25 +102,22 @@ export async function main(event) {
...
 
-  const params = {
+  const updateCaseParams = {
     TableName: config.cases.tableName,
-    Key: {
-      PK: caseKeys.PK,
-      SK: caseKeys.SK,
-    },
-    UpdateExpression: UpdateExpression.join(', '),
+    Key: caseKeys,
```
TeddyAndersson, line 111: Do we need to break out the caseKeys? Feels like it would be easier to read if the key "Key" in updateCaseParams just defines the case keys

```JS
const updateCaseParams = {
  ...other params,
  Key: {
    PK: userCase.PK,
    SK: userCase.SK,
  }
}
```

TeddyAndersson (changes requested):

services/cases-api/lambdas/updateCase.js
TeddyAndersson: In addition to this we probably also want to reset all signatures, if there is a case where several users have signed a case and an update is made that makes the previous case state outdated.

libs/caseStatuses.js
```diff
@@ -11,10 +11,16 @@ const statuses = [
...
+    type: 'active:signature:completed',
     name: 'Signerad',
     description: 'Ansökan är signerad',
   },
+  {
+    type: 'active:signature:pending',
+    name: 'Väntar på signering',
+    description:
+      'Väntar på signering. När båda har signerat med BankID skickas ansökan in för bedömning',
+  },
```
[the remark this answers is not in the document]
dannil76: This is taken from figma... ;)

services/cases-api/lambdas/updateCase.js
dannil76: ...other params?! what is this?

services/cases-api/lambdas/updateCase.js
dannil76: Not sure what to do with this info? Help me understand what to do. :)

KenthLjungHBG: > Not sure what to do with this info? Help me understand what to do. :)

Something like this maybe?

```javascript
const updatedUserCasePeople = userCase.persons?.map(person => {
  const newPerson = { ...person };

  if (newPerson.personalNumber === personalNumber && signature) {
    newPerson.hasSigned = signature.success;
  } else if (answers) {
    newPerson.hasSigned = false;
  }

  return newPerson;
});
```

use-case 1: no answers, no signature (everything stays the same)
use-case 2: answers, no signature (all signatures are reset)
use-case 3: answers, signature (signature is updated for personalNumber provided, all others are reset)
use-case 4: no answers, signature (signature is updated for provided personalNumber only)

dannil76: Done!

dannil76: Fixed

PR #208, opened 2021-07-20
Feat/append encryption attributes

zilaei:

Appending symmetric key attributes when creating a VIVA case with co applicant.

Symmetric key attributes must be present when setting up a Diffie-Hellman shared key between two or more clients:
For EKB, we have encrypted answers that are shared so that the co applicant can approve and sign the case.

## Explain your solution

Appending encryption values to a case if a co applicant is found.
Following values shall be appended when co applicant (with pno 198310011906) present (main applicant pno 196912191118):
```json
encryption: {
      "type": "decrypted",
      "symmetricKeyName": "196912191118:198310011906",
      "primes": {
        "P": 43,
        "G": 10,
      },
      "publicKeys": {
        "196912191118": "",
        "198310011906": "",
      },
    },
```

## How to test the changes?

Concrete example:

1. Checkout this branch.
2. Deploy viva-ms service in your own AWS sandbox environment.
3. In app MH, log in with a user that has a co applicant (ex. user Ali Al-Jabi). Remove any present cases, in AWS dynamoDB cases, before logging in with user.
4. In AWS dynamoDB cases, validate forms[some_id] -> encryption keys and values for the newly created case.

### DIFF
--- a/services/viva-ms/lambdas/createVivaCase.js
+++ b/services/viva-ms/lambdas/createVivaCase.js
@@ -121,28 +121,14 @@ async function putRecurringVivaCase(vivaPerson) {
     endDate: Date.parse(vivaPerson.application.period.end),
   };
 
-  const initialFormAttributes = {
-    answers: [],
-    encryption: { type: 'decrypted' },
-    currentPosition: {
-      currentMainStep: 1,
-      currentMainStepIndex: 0,
-      index: 0,
-      level: 0,
-    },
-  };
-
-  const initialForms = {
-    [recurringFormId]: initialFormAttributes,
-    [completionFormId]: initialFormAttributes,
-  };
+  const formIds = [recurringFormId, completionFormId];
 
   const [getUserError, user] = await to(getUser(PK));
   if (getUserError) {
     throw getUserError;
   }
 
-  const [, formTemplates] = await to(getFormTemplates(initialForms));
+  const [, formTemplates] = await to(getFormTemplates(formIds));
 
   const [getLastUpdatedCaseError, lastUpdatedCase] = await to(
     getLastUpdatedCase(PK, CASE_PROVIDER_VIVA)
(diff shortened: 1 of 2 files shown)


services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -178,6 +178,13 @@ async function putRecurringVivaCase(vivaPerson) {
   const casePersonCoApplicant = getUserByRole(casePersonList, 'coApplicant');
   if (casePersonCoApplicant) {
     caseItemPutParams.Item['GSI1'] = `USER#${casePersonCoApplicant.personalNumber}`;
+
+    const mainApplicantPno = stripNonNumericalCharacters(String(vivaPerson.case.client.pnumber));
+    appendEncryptionAttributes(
+      initialFormAttributes,
+      mainApplicantPno,
```
TeddyAndersson, line 185: mainApplicantPersonalNumber, Pno can be confusing to understand.

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -178,6 +178,13 @@ async function putRecurringVivaCase(vivaPerson) {
   const casePersonCoApplicant = getUserByRole(casePersonList, 'coApplicant');
   if (casePersonCoApplicant) {
     caseItemPutParams.Item['GSI1'] = `USER#${casePersonCoApplicant.personalNumber}`;
+
+    const mainApplicantPno = stripNonNumericalCharacters(String(vivaPerson.case.client.pnumber));
+    appendEncryptionAttributes(
+      initialFormAttributes,
```
TeddyAndersson, line 184: Do not know where initialFormAttributes belongs from, but we should try to deal with this value in an immutable fashion. ie copy the current object and the return a new object.

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -325,3 +332,12 @@ async function getLastUpdatedCase(PK, provider) {
 
   return sortedCases?.[0] || {};
 }
+
+function appendEncryptionAttributes(formAttributes, mainApplicantPno, coApplicantPno) {
```
TeddyAndersson, line 336: argument formAttributes could be any object, so i would rather se this renamed to `object` or `collection` since we do not know if the passed argument is formAttributes or not. We could check this but do not feel that there is any need for this :)

TeddyAndersson (commented):
Left som comment's for improvement in the code :)

KenthLjungHBG (approved):

services/viva-ms/lambdas/createVivaCase.js
```diff
@@ -276,27 +299,33 @@ async function getUser(PK) {
...
   }
 
-  return formTemplates;
+  return forms.reduce(
+    (templates, form) => ({
+      ...templates,
+      [form.Item.id]: form.Item,
+    }),
+    {}
+  );
```
dannil76, line 328: We try to place the return value in descriptive variables before we return.
Suggestion:
formTemplates = forms.reduce...
return formTemplates

One benefit is that you can easy log the variable formTemplates. Good when debugging

The name can certainly be better. Honestly, I have a hard time clearly seeing what the result will be? :)

dannil76 (changes requested):

PR #212, opened 2021-08-25
feat: add log function cu-w8jypq

yo-l1982:
Add a log function to be used as a complement for the api responses with variable info to trace errors etc. in requests.
Will log as json in cloudwatch.

To enable monitoring based on filtering inside logs with consistent log format.
The function should be used for all logs.

## Explain your solution
Basically a wrapper around the [winston](https://www.npmjs.com/package/winston) npm package with some default required variables.

## How to test the changes?

Concrete example:

1. Checkout this branch.
2. Import logger in the services/auth/token/lambdas/generateToken.js and call the logger like the example below.
  ```
import to from 'await-to-js';
import config from '../../../../config';
import secrets from '../../../../libs/secrets';
import * as response from '../../../../libs/response';
import { signToken, verifyToken } from '../../../../libs/token';
import { throwError } from '@helsingborg-stad/npm-api-error-handling';
import tokenValidationSchema from '../helpers/schema';
import { logger } from '../../../../libs/logger';

const CONFIG_AUTH_SECRETS = config.auth.secrets;
const ACCESS_TOKEN_EXPIRES_IN_MINUTES = 20;
const REFRESH_TOKEN_EXPIRES_IN_MINUTES = 30;

export const main = async (event, context) => {
  const [parseJsonError, parsedJson] = await to(parseJson(event.body));
  if (parseJsonError) {
    logger('error', 'Unauthorized!', context.awsRequestId, 'service-auth-token-001');
    return response.failure(parseJsonError);
  }

  const [validationError, validatedEventBody] = await to(
    validateEventBody(parsedJson, tokenValidationSchema)
  );
  if (validationError) {
    logger('error', 'Unauthorized!', context.awsRequestId, 'service-auth-token-002');
    return response.failure(validationError);
  }

  const grantTypeValues = getGrantTypeValues(validatedEventBody);

  const [validateTokenError, decodedGrantToken] = await to(
    validateToken(grantTypeValues.secretsConfig, grantTypeValues.token)
  );
  if (validateTokenError) {
    logger('error', 'Unauthorized!', context.awsRequestId, 'service-auth-token-003');
    return response.failure(validateTokenError);
  }
...
  ```
3. Navigate to the Lambda function in AWS web console and use the "Test" functionality to send some random data to the function.
4. Navigate to the Monitor tab on the Lambda function and select "View logs in Cloudwatch"
5. Find the log and confirm the output.

- [x] Your local serverless enviorment.

### DIFF
--- a/libs/logger.js
+++ b/libs/logger.js
@@ -1,0 +1,22 @@
+import winston from 'winston';
+
+/**
+ * Log function that outputs console logs in json format.
+ *
+ * @param {('error'|'warn'|'info'|'http'|'verbose'|'debug'|'silly')} level Log level based on npm log level standard.
+ * @param {string} message   Log message.
+ * @param {string} requestId Request id to be able to filter logs on a single request.
+ * @param {string} errorCode Error code for locating log in code.
+ * @param {object} data      Custom data related to log.
+ * @returns {void}
+ */
+export const logger = (level, message, requestId, errorCode, data = {}) => {
+  const logger = winston.createLogger({
+    level: process.env.LOG_LEVEL || 'info',
+    format: winston.format.json(),
+    transports: [new winston.transports.Console()],
+    exitOnError: false,
+  });
+
+  logger.log(level, message, { errorCode: errorCode, requestId: requestId, data: data });
+};
--- a/libs/logger.js
+++ b/libs/logger.js
@@ -1,5 +1,12 @@
 import winston from 'winston';
 
+const logger = winston.createLogger({
+  level: process.env.LOG_LEVEL || 'info',
+  format: winston.format.json(),
+  transports: [new winston.transports.Console()],
+  exitOnError: false,
+});
+
 /**
  * Log function that outputs console logs in json format.
  *
@@ -10,13 +17,6 @@ import winston from 'winston';
  * @param {object} data      Custom data related to log.
  * @returns {void}
  */
-export const logger = (level, message, requestId, errorCode, data = {}) => {
-  const logger = winston.createLogger({
-    level: process.env.LOG_LEVEL || 'info',
-    format: winston.format.json(),
-    transports: [new winston.transports.Console()],
-    exitOnError: false,
-  });
-
-  logger.log(level, message, { errorCode: errorCode, requestId: requestId, data: data });
+export const log = (level, message, requestId, errorCode, customData = {}) => {
+  logger.log(level, message, { errorCode: errorCode, requestId: requestId, customData: customData });
 };
--- a/package.json
+++ b/package.json
@@ -29,8 +29,8 @@
   "devDependencies": {
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
-    "@typescript-eslint/parser": "^4.15.0",
     "@types/node": "^14.14.27",
+    "@typescript-eslint/parser": "^4.15.0",
     "babel-eslint": "^10.1.0",
     "commitizen": "^4.2.3",
     "commitlint": "^11.0.0",
@@ -57,7 +57,8 @@
     "camelcase": "^6.2.0",
     "deep-equal": "^2.0.5",
     "hashids": "^2.2.8",
-    "joi": "^17.4.0"
+    "joi": "^17.4.0",
+    "winston": "^3.3.3"
   },
   "config": {
     "commitizen": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -1085,6 +1085,15 @@
   dependencies:
     chalk "^4.0.0"
 
+"@dabh/diagnostics@^2.0.2":
+  version "2.0.2"
+  resolved "https://registry.yarnpkg.com/@dabh/diagnostics/-/diagnostics-2.0.2.tgz#290d08f7b381b8f94607dc8f471a12c675f9db31"
+  integrity sha512-+A1YivoVDNNVCdfozHSR8v/jyuuLTMXwjWuxPFlFlUapXoGc+Gj9mDlTDDfrwl7rXCl2tNZ0kE8sIBO6YOn96Q==
+  dependencies:
+    colorspace "1.1.x"
+    enabled "2.0.x"
+    kuler "^2.0.0"
+
 "@eslint/eslintrc@^0.4.1":
   version "0.4.1"
   resolved "https://registry.yarnpkg.com/@eslint/eslintrc/-/eslintrc-0.4.1.tgz#442763b88cecbe3ee0ec7ca6d6dd6168550cbf14"
@@ -2326,6 +2335,11 @@ async@^2.6.2:
   dependencies:
     lodash "^4.17.14"
 
+async@^3.1.0:
+  version "3.2.1"
+  resolved "https://registry.yarnpkg.com/async/-/async-3.2.1.tgz#d3274ec66d107a47476a4c49136aacdb00665fc8"
+  integrity sha512-XdD5lRO/87udXCMC9meWdYiR+Nq6ZjUfXidViUZGu2F1MO4T3XwZ1et0hb2++BgLfhyJwy44BGB/yx80ABx8hg==
+
 async@^3.2.0:
   version "3.2.0"
   resolved "https://registry.yarnpkg.com/async/-/async-3.2.0.tgz#b3a2685c5ebb641d3de02d161002c60fc9f85720"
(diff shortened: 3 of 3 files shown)


helsingborgsstad:
Task linked: [CU-w8jypq Monitorering: API Konsistent loggformat. Skapa generell funktion för loggning.](https://app.clickup.com/t/w8jypq)

TeddyAndersson (approved):
The function works like a charm, had no issues with outputting the logging by following the test description.

Had som thoughts about the function comment, since we are trying to avoid comments in the code base and instead write more declarative code. But for me this comment is somewhere in between since it fullfills the purpose of exlpaning the params and the overall intention of the function.

The longterm issue might be that the function gets updated and the comment is not and thereby the comment will be misleading.

libs/logger.js
```diff
@@ -0,0 +1,22 @@
...
+ * Log function that outputs console logs in json format.
+ *
+ * @param {('error'|'warn'|'info'|'http'|'verbose'|'debug'|'silly')} level Log level based on npm log level standard.
+ * @param {string} message   Log message.
+ * @param {string} requestId Request id to be able to filter logs on a single request.
+ * @param {string} errorCode Error code for locating log in code.
+ * @param {object} data      Custom data related to log.
+ * @returns {void}
+ */
+export const logger = (level, message, requestId, errorCode, data = {}) => {
```
KenthLjungHBG, line 13: Calling it `logger` seems off to me, as nouns are usually reserved for classes etc. I would suggest renaming it to a verb like `log` (but maybe that's too short/conflicting) 🙂.

libs/logger.js
```diff
@@ -0,0 +1,22 @@
...
+  const logger = winston.createLogger({
+    level: process.env.LOG_LEVEL || 'info',
+    format: winston.format.json(),
+    transports: [new winston.transports.Console()],
+    exitOnError: false,
+  });
```
KenthLjungHBG, line 19: Perhaps a small optimization in practice but could we cache this somehow for the duration of the lambda execution? This is (re)initialized for every log message.

KenthLjungHBG (commented):
Works nicely, however would it perhaps be a good idea to also provide wrapper functions for common log levels? Like `logInfo`, `logError` etc.? It would clean up the usage a bit and also reduce accidents such as mistyping `eror`.

KenthLjungHBG:
> The function works like a charm, had no issues with outputting the logging by following the test description.
> 
> Had som thoughts about the function comment, since we are trying to avoid comments in the code base and instead write more declarative code. But for me this comment is somewhere in between since it fullfills the purpose of exlpaning the params and the overall intention of the function.
> 
> The longterm issue might be that the function gets updated and the comment is not and thereby the comment will be misleading.

For me these kind of documentation comments really helps to get a overview of all the details of a function without having to dig into its implementation 🙂. One issue might be that it becomes hard after a while to sift through if half a file is just comments, but overall I think it's better to have than not to. IDEs usually provide ample functionality to fold/hide comments if necessary, and having comments helps intellisense a lot (for example while testing now I got proper suggestion for all values of log level).

I also agree that there is an issue in case code changes and documentation does not, however I also feel that that is the responsibility of the developer that changes to code to also update the comments, as well as the responsibility of the rest of the team to catch 🙂.

dannil76:
> > The function works like a charm, had no issues with outputting the logging by following the test description.
> > Had som thoughts about the function comment, since we are trying to avoid comments in the code base and instead write more declarative code. But for me this comment is somewhere in between since it fullfills the purpose of exlpaning the params and the overall intention of the function.
> > The longterm issue might be that the function gets updated and the comment is not and thereby the comment will be misleading.
> 
> For me these kind of documentation comments really helps to get a overview of all the details of a function without having to dig into its implementation 🙂. One issue might be that it becomes hard after a while to sift through if half a file is just comments, but overall I think it's better to have than not to. IDEs usually provide ample functionality to fold/hide comments if necessary, and having comments helps intellisense a lot (for example while testing now I got proper suggestion for all values of log level).
> 
> I also agree that there is an issue in case code changes and documentation does not, however I also feel that that is the responsibility of the developer that changes to code to also update the comments, as well as the responsibility of the rest of the team to catch 🙂.

Please watch this clip and then we maybe can agree on something in the middle
https://youtu.be/2a_ytyt9sf8

libs/logger.js
yo-l1982: Agree, I could just move it outside the function like this.

Looks a bit unnessesary with a seperate funtion but my reasoning about having it as a seperate function and now just plain winston was to push devs to always fill in the info needed in monitoring.

```js
import winston from 'winston';

const logger = winston.createLogger({
  level: process.env.LOG_LEVEL || 'info',
  format: winston.format.json(),
  transports: [new winston.transports.Console()],
  exitOnError: false,
});

/**
 * Log function that outputs console logs in json format.
 *
 * @param {('error'|'warn'|'info'|'http'|'verbose'|'debug'|'silly')} level Log level based on npm log level standard.
 * @param {string} message   Log message.
 * @param {string} requestId Request id to be able to filter logs on a single request.
 * @param {string} errorCode Error code for locating log in code.
 * @param {object} data      Custom data related to log.
 * @returns {void}
 */
export const log = (level, message, requestId, errorCode, data = {}) => {
  logger.log(level, message, { errorCode: errorCode, requestId: requestId, data: data });
};
```

libs/logger.js
yo-l1982: will do

libs/logger.js
```diff
@@ -0,0 +1,22 @@
+import winston from 'winston';
+
+/**
+ * Log function that outputs console logs in json format.
+ *
+ * @param {('error'|'warn'|'info'|'http'|'verbose'|'debug'|'silly')} level Log level based on npm log level standard.
+ * @param {string} message   Log message.
```
dannil76, line 7: Can be removed.

libs/logger.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {string} requestId Request id to be able to filter logs on a single request.
```
dannil76, line 8: Can be removed.

libs/logger.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {string} errorCode Error code for locating log in code.
```
dannil76, line 9: Can be removed.

libs/logger.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param {object} data      Custom data related to log.
```
dannil76, line 10: Comment suggest the param can be named customData?

dannil76 (changes requested):

libs/logger.js
yo-l1982: Can do that.

dannil76 (approved):

dannil76:
> I can say for sure the function will be harder to understand without the function comments.
> Suggestions how to make it easier to understand would be appreciated

dannil76:
Sorry. Not sure what I just did? :)

KenthLjungHBG:
> Sorry. Not sure what I just did? :)

Looks like a "close with comment" but seems you fixed it immediately

libs/logger.js
dannil76: or event customLogData. Then again all ting is data so I would exclude it from the name and maybe just name it customLog

dannil76:
@param {string} message   Log message.
I think this is self-explanatory. Then you could name the param logMessage to be even more clear.

dannil76:
> > Sorry. Not sure what I just did? :)
> 
> Looks like a "close with comment" but seems you fixed it immediately

I sure hope so :)

KenthLjungHBG:
> @param {string} message Log message.
> I think this is self-explanatory. Then you could name the param logMessage to be even more clear.

I think naming the variable `logMessage` would be redundant since the `log` part is implied from the function.

KenthLjungHBG:
> > > The function works like a charm, had no issues with outputting the logging by following the test description.
> > > Had som thoughts about the function comment, since we are trying to avoid comments in the code base and instead write more declarative code. But for me this comment is somewhere in between since it fullfills the purpose of exlpaning the params and the overall intention of the function.
> > > The longterm issue might be that the function gets updated and the comment is not and thereby the comment will be misleading.
> > 
> > 
> > For me these kind of documentation comments really helps to get a overview of all the details of a function without having to dig into its implementation 🙂. One issue might be that it becomes hard after a while to sift through if half a file is just comments, but overall I think it's better to have than not to. IDEs usually provide ample functionality to fold/hide comments if necessary, and having comments helps intellisense a lot (for example while testing now I got proper suggestion for all values of log level).
> > I also agree that there is an issue in case code changes and documentation does not, however I also feel that that is the responsibility of the developer that changes to code to also update the comments, as well as the responsibility of the rest of the team to catch 🙂.
> 
> Please watch this clip and then we maybe can agree on something in the middle
> https://youtu.be/2a_ytyt9sf8

I have watched the video now; very interesting perspective 🙂. However I would argue that having a JSDoc-like comment is still useful here due to the fact that JavaScript is not typed, so with just proper naming alone it would be hard to infer the intended input types and return values. This being an API-level generic function I think is a good place to have a function comment for future outside collaborators to take part of.

I also agree that having comments like "@param {string} message Log message." is redundant, however I'm not sure it would be proper to remove some params but keep others (the log level param being an obvious keeper for me). Not sure what works best here.

TeddyAndersson:
Created a card https://app.clickup.com/t/1dndpeu for this in the backlog so that we can get som work done by implementing code standards. Let's take this ongoing discussion and do some work to avoid these discussions in the future.

For me this code looks and works just fine, functional comment or not

PR #213, opened 2021-08-28
Fix/navet ms error respons

dannil76:
Fix error exception handling. Overall refactoring to increase code readability

See Clickup task for more info


helsingborgsstad:
Task linked: [CU-yrpwbr Navet poll - Om certifikat är ogiltigt kastas ett underfined TypeError](https://app.clickup.com/t/yrpwbr)

services/navet-ms/helpers/parser.js
```diff
@@ -1,4 +1,5 @@
 import parser from 'xml2js';
+import { ResourceNotFoundError } from '@helsingborg-stad/npm-api-error-handling';
 
 export const parseXml =
```
KenthLjungHBG, line 4: While reading through this for the first time I found `parseXml` and `parseJSON` to be a bit misleading. `parseXml` doesn't actually parse any XML, it takes an object and constructs a specific XML payload. Maybe something like `makePersonpostRequestXML` would be more appropriate?

`parseJSON` doesn't parse JSON either, it takes XML and parses it into a JSON object. A bit more unsure about this one, maybe `parseToJSON` or `parseXML`? Considering the inconsistencies with the 3rd party function `parser.parseString` following the `parseX` convention where X denotes the input type. On the other hand, functions like `parseInt` follow a different pattern where X denotes the output type, but I'm mentioning it since I found it confusing at first.

These are just some side-suggestions while we're editing this file anyway 🙂.

services/navet-ms/serverless.yml
```diff
@@ -17,7 +16,6 @@ provider:
   region: eu-north-1
   endpointType: regional
   tracing:
-    apiGateway: true
```
KenthLjungHBG, line 20: I'm not sure what `tracing` does but isn't it useful for debugging? Why is `apiGateway` irrelevant? Is it because `navet-ms` is never triggered by API requests, only `bankid.collect`? Does it affect CloudWatch/logging?

services/navet-ms/helpers/parser.js
```diff
@@ -18,20 +19,26 @@ export const parseJSON = input =>
...
-        const options = {
-          trim: true,
-          explicitArray: false,
-        };
-        parser.parseString(parsedTwice[0], options, (error, result) => {
-          if (error) throw error;
-          resolve(result);
-        });
+
+      // If result has any posts the length of posts will more than 2
```
TeddyAndersson, line 23: This comment is for dumb people remove

TeddyAndersson: Jokes aside <3

services/navet-ms/helpers/parser.js
```diff
@@ -18,20 +19,26 @@ export const parseJSON = input =>
...
+      if (posts.length < 2) {
+        throw new ResourceNotFoundError();
       }
-      resolve(input);
+
+      const parsedTwice = posts[1].split('</ns0:');
```
TeddyAndersson, line 28: parsedTwice, what is parsed twice?

services/navet-ms/lambdas/pollUser.js
```diff
@@ -1,78 +1,106 @@
...
-    firstName: navetData.Namn.Fornamn,
-    lastName: navetData.Namn.Efternamn,
+
+  const [requestNavetUserError, navetUser] = await to(requestNavetUser(user.personalNumber));
+  if (requestNavetUserError) {
+    return console.error('(Navet-ms)', requestNavetUserError);
+  }
+
+  const eventDetail = getNavetPollEventDetail(navetUser);
+  await putEvent(eventDetail, 'NavetPoll', 'navet.poll');
```
TeddyAndersson, line 23: Suggestion for improvement, have a putNavetPollEvent function, that has the scope of creating and putting the event tot the event bridge.

services/navet-ms/lambdas/pollUser.js
```diff
@@ -1,78 +1,106 @@
...
+      throw navetResponseXmlErrorMessage;
+    }
+
+    throw postNavetClientError;
   }
 
-  // Parse XML data to JSON
-  [err, parsedData] = await to(parseJSON(navetUser.data));
-  if (err) throwError(500);
+  return await getNavetUserPersonPostAsJson(navetUser);
```
TeddyAndersson, line 75: getNavetPersonPost is a bit more concise for and easier to read. Do we need to know that the return value is Json?

services/navet-ms/lambdas/pollUser.js
```diff
@@ -1,78 +1,106 @@
...
+}
 
-  const { Folkbokforingspost } = parsedData;
+async function getNavetUserPersonPostAsJson(navetUser) {
+  const [parseJsonError, navetUserJson] = await to(parseJSON(navetUser.data));
```
TeddyAndersson, line 79: The parseJSON function does not parse json at all. So very misleading, what this function (parseJSON) does is that it converts xml to json. Therefore convertXmlToJson, toJson or xmlToJson would be more suitable naming options.

TeddyAndersson: Another note is also that we simply could return the xml as json and thereby have a json representation of the xml data then let another function handle the retrieval of posts/post.

```JS
const navetPerson = xmlToJson(xml) => returns xml as json
const navetPersonPost = getPost(navetPerson)
```

services/navet-ms/serverless.yml
TeddyAndersson: You are right @KenthLjungHBG the navet-ms is not related to the apiGateway. Do not think this will change anything at all for logging purposes, but not 100% sure

services/navet-ms/lambdas/pollUser.js
```diff
@@ -1,78 +1,106 @@
...
+    throw new UnauthorizedError();
   }
 
-  // Collect the user data from response
-  const navetUserData = Folkbokforingspost.Personpost;
-  if (navetUserData === undefined) throwError(500);
+  return Folkbokforingspost.Personpost;
+}
+
+function isUserConfidential(navetUser) {
```
TeddyAndersson, line 96: isNavetPersonCondifential seems more correct since we don't want to this to be missunderstud as a aws user.

TeddyAndersson (changes requested):
Some minor changes to be corrected.

services/navet-ms/lambdas/pollUser.js
KenthLjungHBG: A note that I just realised is that JSON technically speaking is the string format. The correct type `parseJSON` returns is a full JavaScript Object; it doesn't really directly relate to JSON specifically.

TeddyAndersson: You are right might be good to consider this when making changes since we most of the time are handling a JS Object and not JSON. JSON is rather used in return statements towards a client or in a request towards a underlaying business system.

services/navet-ms/serverless.yml
dannil76: True @TeddyAndersson

services/navet-ms/helpers/parser.js
dannil76: Do it then

services/navet-ms/helpers/parser.js
```diff
@@ -5,8 +5,8 @@ export const getPersonPostSoapRequest = ({
   orderNumber,
   organisationNumber,
   personalNumber,
-  xmlEvnUrl,
-}) => `<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:v1="${xmlEvnUrl}">
+  xmlEnvUrl,
+}) => `<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:v1="${xmlEnvUrl}">
```
dannil76: Not sure what is changed?

TeddyAndersson: Changed the naming of xmlEvnUrl to xmlEnvUrl  (Evn => Env)( open for suggestions of the naming )

PR #215, opened 2021-09-01
Fix/user create event flow

dannil76:
## Feature description
When the user logs in via the app for the first time the user is created in the users table (DynamoDb).
Viva-ms needs to ensure that the user indeed is created before it can create a new (open) application period.

## Solution description
New lambda - users-ms/findUser is added to check if logged in user exists in the users table.
If no - findUser emits an event to inform that the user could not be found.

The lambda - users-ms/createUser is configured to trigger if findUser does not find the user and will add the user to the users table.
createUser emits an event to inform that the user is added in the users table.

findUser is configured to trigger if a user has been added in the users table.
findUser is executed once again and will find the user.

If yes - findUser emits an event to inform that the user exists.
viva-ms/createVivaCase is configured to trigger on the user exists event emitted by users-ms/findUser

_OBS! syncUser does not exist at this point, so don't mind that part of the flowchart :)_
image

users-ms/findUser, users-ms/createUser, viva-ms/createVivaCase. serverless.yml is altered in order to create the necessary event flow and logic

## How to test
Check out this branch.
Redeploy services users-ms and viva-ms.

Test 1
Delete the user Ylva Jansson (195809262743) from the users table
Delete all cases related to Ylva in the cases table
Trigger bankId.collect by executing the script found in mock/putevent.sh

Test 2
Delete all cases related to Ylva in the cases table
Trigger bankId.collect once again to ensure that a new case is created for Ylva

The expected result should follow the flow explained above.

One more thing. There need to be an open period for application for Ylva in Viva-admin.
At the time of writing, there is a period open for application. May disappear, however, as Maria tests.

### DIFF
--- a/services/users-ms/lambdas/findUser.js
+++ b/services/users-ms/lambdas/findUser.js
@@ -1,5 +1,4 @@
 /* eslint-disable no-console */
-import * as errorHandler from '@helsingborg-stad/npm-api-error-handling';
 import to from 'await-to-js';
 
 import config from '../../../config';
@@ -16,9 +15,9 @@ export async function main(event) {
   }
 
   if (userItem == undefined) {
-    const [emitError] = await to(emitEventUserNotFound(userDetail));
-    if (emitError) {
-      throw new InternalServerError(emitError);
+    const [emitEventError] = await to(emitEventUserNotFound(userDetail));
+    if (emitEventError) {
+      return console.error('(users-ms) emitEventError', emitEventError);
     }
     return console.log(
       `User with personal number: ${personalNumber}, could not be found in the users table.`
@@ -39,7 +38,7 @@ async function getUser(personalNumber) {
 
   const [getError, getResult] = await to(dynamoDB.call('get', params));
   if (getError) {
-    throw new errorHandler.InternalServerError(getError);
+    throw getError;
   }
 
   return getResult.Item;
--- a/services/users-ms/lambdas/findUser.js
+++ b/services/users-ms/lambdas/findUser.js
@@ -11,13 +11,13 @@ export async function main(event) {
 
   const [getUserError, userItem] = await to(getUser(personalNumber));
   if (getUserError) {
-    return console.error('(users-ms) getUserError', getUserError);
+    return console.error('(users-ms: findUser) getUserError', getUserError);
   }
 
   if (userItem == undefined) {
     const [emitEventError] = await to(emitEventUserNotFound(userDetail));
     if (emitEventError) {
-      return console.error('(users-ms) emitEventError', emitEventError);
+      return console.error('(users-ms: findUser) emitEventError', emitEventError);
     }
     return console.log(
       `User with personal number: ${personalNumber}, could not be found in the users table.`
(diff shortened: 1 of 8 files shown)


helsingborgsstad:
Task linked: [CU-nzdyk6 När sökande loggar in för första gången visas inga ärenden](https://app.clickup.com/t/nzdyk6)

services/users-ms/lambdas/findUser.js
```diff
@@ -0,0 +1,55 @@
...
+  const userDetail = event.detail;
+  const { personalNumber } = userDetail;
+
+  const [getUserError, userItem] = await to(getUser(personalNumber));
+  if (getUserError) {
+    return console.error('(users-ms: findUser) getUserError', getUserError);
+  }
+
+  if (userItem == undefined) {
+    const [emitEventError] = await to(emitEventUserNotFound(userDetail));
```
TeddyAndersson, line 18: To keep the code consistent i would suggest to build on the putEvent naming convention, emitEvent is a new naming for the same functionality. This made me a bit confused, due to the knowledge of our putEvent function from the libs.

putEvent in itself might not be the best name for this function but it is what is used currently, so would be good if we could stick to that for know :)

services/users-ms/lambdas/findUser.js
```diff
@@ -0,0 +1,55 @@
...
+  return getResult.Item;
+}
+
+async function emitEventUserNotFound(user) {
+  return putEvent(user, 'usersMsFindUserUnsucceeded', 'usersMs.findUser');
+}
+
+async function emitEventUserExists(user) {
+  return putEvent(user, 'usersMsFindUserSuccess', 'usersMs.findUser');
+}
```
TeddyAndersson, line 55: These functions feels very clear in their purpose but at the same time redundant, do not have any suggestions for other solutions at this moment but there is room for improvement i guess :)

services/users-ms/lambdas/findUser.js
```diff
@@ -0,0 +1,55 @@
...
+
+  const [emitEventError] = await to(emitEventUserExists(userItem));
+  if (emitEventError) {
+    return console.error('(users-ms: findUser) emitEventError', emitEventError);
+  }
+  console.info('User was found in the users table.');
+  return true;
+}
+
+async function getUser(personalNumber) {
```
TeddyAndersson, line 33: This code block feels familiar, are we not doing this action towards dynamoDb elsewhere? If so we have multiple functions in different places that has the same task :)

TeddyAndersson: Ie we then have redundancy in the code.

TeddyAndersson (changes requested):

services/users-ms/lambdas/findUser.js
TeddyAndersson: A thought on these functions might be to declare them in an object

```JS
const userEvents = {
  exists: async (user) => putEvent(user, ...),
  notFound: async (user) => putEvent(user, ...)
}

// then we use them like this
const eventResponse = userEvent.exists(user)
const eventResponse = userEvent.notFound(user)
```
You could also go further and do something like
```JS
// extend the incoming user object with functions/methods
const user = {
  ...user, 
  event: {
    send: (type) => putUserEvent(user, type) // don't now if there is a better way of referring to the user object here, but you get my intention
  }
}

async function putUserEvent(user, type) {
  const eventTypeCollection = {
   notFound: {...},
   exists: {...},
  }

  return await putEvent(...eventTypeCollection[type])
}

user.event.put('exists') 
```

TeddyAndersson (approved):

PR #216, opened 2021-09-01
Feat/logging cu w8k9e4

yo-l1982:
WIP - Not fullly tested

Replaces log output everywhere.

Changed log function after feedback in previous PR.
Specific log level functions and renaming.

Preferably we want to keep only "error" log level in production(unless tracking something critical) and it would hide all the other log levels logged.
Log levels below error only to be used in testing/dev environments.

So to reduce the amount of clutter-logs some should maybe moved to "warn" or "info".


helsingborgsstad:
Task linked: [CU-w8k9e4 Monitorering: API Konsistent loggformat. Använd ny logg-funktion på alla fel.](https://app.clickup.com/t/w8k9e4)

services/auth/token/lambdas/generateToken.js
```diff
@@ -5,21 +5,39 @@ import * as response from '../../../../libs/response';
...
   const [parseJsonError, parsedJson] = await to(parseJson(event.body));
+
   if (parseJsonError) {
+    logWarn(
+      'JSON Parse error',
+      context.awsRequestId,
+      'service-auth-token-generateToken-001',
+      parseJsonError
+    );
+
```
TeddyAndersson, line 25: These function calls to the log function makes it harder to read the overall code, would prefer if these calls could be boiled down to oneliners instead. 

A suggestion here is to strip the static parameters into a object/collection where they can be retrieved and then passed them to the log function.
```JS
const logs = {
  jsonParseError: {
    message: 'JSON parse error',
    requestId: context.awsRequestId,
    errorCode: 'service-auth-token-generateToken-001'
  }
}

logWarn({ ...logs.jsonParseError, customData: parseJsonError })

```

TeddyAndersson: From what i can see the last parameter for customData is a bit more dynamic, so that's way in this case i did not declare it in the logs object

services/auth/token/lambdas/generateToken.js
```diff
@@ -5,21 +5,39 @@ import * as response from '../../../../libs/response';
...
 
   const [validationError, validatedEventBody] = await to(
     validateEventBody(parsedJson, tokenValidationSchema)
   );
+
   if (validationError) {
+    logWarn(
+      'Validation error',
+      context.awsRequestId,
+      'service-auth-token-generateToken-002',
```
TeddyAndersson, line 37: Also curios on the thought behind the errorCode, what is the naming convention based upon and what's the purpose? easier to find/search for in the logs?

yo-l1982: Its generally used to distinguish error logs with the same error message to make it easy to find in code.
The convention here is a bit over the top with this template ${folder-names}-${filename}-${unique-number}

libs/logs.js
```diff
@@ -0,0 +1,36 @@
+import winston from 'winston';
+
+const logger = winston.createLogger({
+  level: process.env.LOG_LEVEL || 'info',
+  format: winston.format.json(),
+  transports: [new winston.transports.Console()],
+  exitOnError: false,
+});
+
+export const log = (level, message, requestId, errorCode, customData = {}) => {
```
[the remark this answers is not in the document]
KenthLjungHBG: I like this idea, however that would make some parameters optional. Maybe at least provide message as well, so `log(level, message, options)`? Also maybe name it `extraParams` since I think `options` implies that it's being used to modify the code behavior 🙂.

services/auth/token/lambdas/generateToken.js
yo-l1982: That is correct.
I will look into automating it. Would make it easier to use, this would probably be a typical copy paste code snippet and it might be easy to forget to change or missunderstand its format/purpose.

services/auth/token/lambdas/generateToken.js
```diff
@@ -5,21 +5,39 @@ import * as response from '../../../../libs/response';
...
 import tokenValidationSchema from '../helpers/schema';
+import { logWarn } from '../../../../libs/logs';
 
 const CONFIG_AUTH_SECRETS = config.auth.secrets;
 const ACCESS_TOKEN_EXPIRES_IN_MINUTES = 20;
 const REFRESH_TOKEN_EXPIRES_IN_MINUTES = 30;
 
-export const main = async event => {
+export const main = async (event, context) => {
+  console.log(event);
```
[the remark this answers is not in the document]
yo-l1982: Me being messy. will remove.

libs/logs.js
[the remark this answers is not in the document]
TeddyAndersson: I guess the second parameter is up for discussion in regards of the name, in my opinon neither ´options´ or ´extraParams´ are good variabel names for these extra params. We should have a name that indicates the purpose of the parameters so that it's clear for the person reading the code (y)

services/auth/token/lambdas/authorize.js
```diff
@@ -4,12 +4,20 @@ import generateIAMPolicy from '../helpers/generateIAMPolicy';
...
 const CONFIG_AUTH_SECRETS_ACCESS_TOKEN = config.auth.secrets.accessToken;
 
-export async function main(event) {
+export async function main(event, context) {
   const { authorizationToken } = event;
 
+  if (!authorizationToken) {
+    logWarn('Unauthorized!', context.awsRequestId, 'service-auth-token-authorize-001');
+
+    throwError(401, 'Unauthorized');
```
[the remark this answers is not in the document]
yo-l1982: I see, I will revert that

libs/logs.js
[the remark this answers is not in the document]
yo-l1982: I kind of want the requestId required.
Concurrent requests to lambdas makes it hard to follow logs without the requestId as a filter, then I rather make the message optional.
Is it more "messy" if I validate the object input?

libs/logs.js
```diff
@@ -0,0 +1,36 @@
...
+  log('info', message, requestId, errorCode, customData);
+};
+
+export const logVerbose = (message, requestId, errorCode, customData = {}) => {
+  log('verbose', message, requestId, errorCode, customData);
+};
+
+export const logDebug = (message, requestId, errorCode, customData = {}) => {
+  log('debug', message, requestId, errorCode, customData);
+};
```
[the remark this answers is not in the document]
yo-l1982: Ill do this!

services/auth/token/lambdas/generateToken.js
yo-l1982: hmm this is dependent on the above comment about using object as parameters right?

services/cases-api/lambdas/updateCase.js
```diff
@@ -17,38 +17,54 @@ import { getStatusByType } from '../../../libs/caseStatuses';
...
+    logError(errorMessage, context.awsRequestId, 'service-cases-api-updateCase-001');
+
+    return response.failure(new BadRequestError(errorMessage));
   }
 
   const { error, value: validatedJsonBody } = updateCaseValidationSchema.validate(requestJsonBody, {
     abortEarly: false,
   });
   if (error) {
+    logError(error.message, context.awsRequestId, 'service-cases-api-updateCase-002', error);
```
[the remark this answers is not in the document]
yo-l1982: fixed

services/pdf-ms/lambdas/addPdfToCase.ts
```diff
@@ -17,12 +17,14 @@ import {
...
 
+import log from '../../../libs/logs';
+
 const PDF_SSM_PARAMS = params.read(config.pdf.envsKeyName);
 
 // Convert DynamoDB item to plain object
 const dynamoDbConverter = AWS.DynamoDB.Converter;
 
-export async function main(event: Record<string, any>): Promise<Boolean> {
+export async function main(event: Record<string, any>, context): Promise<Boolean> {
```
dannil76, line 27: To solve this code issue you could add **Record<string, any>** after context.

dannil76 (changes requested):

services/pdf-ms/lambdas/addPdfToCase.ts
yo-l1982: Added.

dannil76 (approved):

services/auth/token/lambdas/generateToken.js
KenthLjungHBG: Yes, so this feels a bit convoluted. I can't think of any way that would be much cleaner than how it is right now.

yo-l1982: Ok Ill keep it as is for now.

PR #218, opened 2021-09-09
Fix/bankid collect user agent check

TeddyAndersson:
The BankID collect lambda have a new check in place to verify if the incoming request is from our Mitt Helsingborg Application (iOS/Android). 

If the this requirement is fulfilled we (a) send a bankid.collect event to the AWS event-bridge and (b) pass a new created authorization-code in the json response back to the client.

If the requirement is not full-filled we only return BankID related information to the client.

## Explain your solution

The solution for verifying if a request comes from our trusted Mitt Helsingborg app is to check the User-Agent header that is passed in the request towards the bankid/collect endpoint. If the User-Agent string contains `MittHelsingborg` we fire the event and return an authorization-code alongside with bankid related information. 
## How to test the changes?

In order to test these changes you need to do the following:

(1) Deploy the BankID service in your sandbox. 
(2) Make sure you have a user that is setup with bankid on your computer. (
3) Start a bankid auth process by sending a request towards the bankid/auth endpoint and copy the orderRef in the response. 
(4) Send a request towards the bankid/collect endpoint with the previous copied orderRef to check that bankid order is pending. 
(5) Verify the pending order request with bankid on your computer. 
(6) Check the response from bankid/collect it should not contain an authorizationCode attribute.
(7) Now repeat step 3 to 5 but on step 4 pass a User-Agent header that contains MittHelsingborg.
(8) Check the response and it should contain an authorizationCode attribute.
(9) Check the logs in your aws console to see if the navet-poll lambda have been triggered. (Should only happen if the User-Agen header is passed and contains MittHelsingborg)

- [X] AWS personal sandbox.
- [X] AWS personal sandbox + Mitt Helsingborg iOS application
- [] AWS personal sandbox + Mitt Helsingborg Android application

## Additional Notes
This is a patch to avoid creation of users when third party clients uses our bankid service. This is however not a solid solution and needs an overhaul of the auth in conjunction with bankid should be made. The bankid service and suth service are not loose-coupled at the moment, and they really should be in the future.

### DIFF
--- a/services/bankid-api/lambdas/collect.js
+++ b/services/bankid-api/lambdas/collect.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
 import { throwError } from '@helsingborg-stad/npm-api-error-handling';
 import to from 'await-to-js';
 import config from '../../../config';
@@ -17,7 +18,6 @@ export const main = async event => {
   const { orderRef } = JSON.parse(body);
   const bankidSSMParams = await SSMParams;
 
-  // eslint-disable-next-line no-console
   console.info('🚀 ~ file: collect.js ~ line 21 ~ headers -> User-Agent', headers['User-Agent']);
 
   const payload = { orderRef };
@@ -66,7 +66,6 @@ export const main = async event => {
     };
   }
 
-  // eslint-disable-next-line no-console
   console.info('🚀 ~ file: collect.js ~ line 70 ~ responseAttributes', responseAttributes);
 
   return response.success(200, {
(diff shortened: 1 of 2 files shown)


helsingborgsstad:
Task linked: [CU-1e6u03k Användare skapas upp från andra tjänster som använder bankid (kulturkortet)](https://app.clickup.com/t/1e6u03k)

dannil76 (approved):

dannil76:
Can we trust that User-Agent is always MittHelsingborg. Without spaces?

PR #220, opened 2021-09-17
feat: setting up booking-api

karlknowit:
## What was solved
Added a new service (booking-api) for making requests against Datatorget to handle outlook calendar reqests (create, update, get and cancel/delete).

## How was it solved
Solved it by creating a new serverless stack containing lambda functions (one for each respective task) which makes queries against Datatorget.

This structure follow the agreed way of creating serverless stacks in the repository.

## How was it tested
Tested it by deploying the stack and making successful requests against all Api Gateway paths with postman.


services/booking-api/helpers/makeBookingRequest.js
```diff
@@ -0,0 +1,15 @@
...
+      { rejectUnauthorized: false },
+      { 'X-ApiKey': apikey }
+    );
+
+    const response = await request.call(requestClient, 'post', url, body);
+    return response;
+  } catch (error) {
+    throw new Error(error);
+  }
+};
```
TeddyAndersson, line 15: I think you can skip the try-catch block here and instead use the await-to-js to handle the promise and throw error if it occurs.

```JS
async function makeBookingRequest(url, apiKey, body) {
   const requestClient = request.requestClient(
        { rejectUnauthorized: false },
        { 'X-ApiKey': apikey }
    );

    const [response, error] = await to(request.call(requestClient, 'post', url, body));
    if (error) {
       throw error
    }
    return response;
}
```

TeddyAndersson: Would also consider renaming the function/name to something mor explicit. A name that represents the intention or expected outcome for the function, ie what form of request is this? Is it a creation of a singel booking or multiple ones, or something else? :)

services/booking-api/babel.config.js
```diff
@@ -0,0 +1,5 @@
+const babelBase = require('../../babel-base.config');
+
+module.exports = {
+  ...babelBase,
+};
```
TeddyAndersson, line 5: What is the prupose of adding this configuration, is it in order for jest to function? :)

services/booking-api/helpers/getSsmParameters.js
```diff
@@ -0,0 +1,16 @@
+import params from '../../../libs/params';
+import config from '../../../config';
+
+let bookingSsmParameters;
+
+export default async () => {
```
TeddyAndersson, line 6: Same goes here for function declaration as previously mentioned :)

TeddyAndersson (changes requested):
The code looks good overall, have some minor changes in order for the code base to not deviate to much. But great work (y)

services/booking-api/helpers/makeBookingRequest.js
karlknowit: Having a hard time to naming this function to a more specific one. Since this function is making requests against Datatorget with multiple outcomes, i.e. creating/cancel/get bookings. So the outcome differs depending on which url and body that it receives.

services/booking-api/lambdas/cancel.js
```diff
@@ -0,0 +1,23 @@
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
+import to from 'await-to-js';
+
+import * as response from '../../../libs/response';
+
+import getSsmParameters from '../helpers/getSsmParameters';
+import makeBookingRequest from '../helpers/makeBookingRequest';
+
+export const main = async event => {
```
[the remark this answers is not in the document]
karlknowit: Sounds reasonable, will change

services/booking-api/helpers/makeBookingRequest.js
TeddyAndersson: Then i think sendBookingPostRequest is sufficient maybe? Or simply have helper functions that use this function ie sendCancelBookingRequest, sendCreateBookingRequest etc :)

You could look at the viva-ms/helpers/vivaRequestClient for some inspiration in how to setup this and export the requests towards the datatorget api.

karlknowit: Let's start by only changing the name to a little more specific one. If we don't think that's enough in the future then I suggest we change it then and keep it as simple as possible for now.

services/booking-api/README.md
```diff
@@ -0,0 +1,198 @@
+# HELSINGBORG IO SLS BOOKING SERVICE
```
TeddyAndersson, line 1: Might be a good idea to add instructions for testing in the readme file (y)

services/booking-api/lambdas/cancel.js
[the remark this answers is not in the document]
TeddyAndersson: In addition to this it's perfectly fine to use arrow declaration in a function to assign a value. 💃 
```JS
function example(params) {
  const nestedExample = (params) => {...} // this function declaration should not be to extensive though, then refactoring makes more sense :) 
  // use nestedExample var to something 
}
```

KenthLjungHBG (changes requested):

services/booking-api/package.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "booking-api",
+  "version": "0.0.1",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "node ../../node_modules/.bin/jest"
```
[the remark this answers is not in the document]
KenthLjungHBG: Nevermind this, I didn't pay attention to the structure 🤦‍♂️.

package.json
```diff
@@ -40,6 +43,8 @@
     "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-prettier": "^3.3.1",
     "husky": "^4.3.8",
+    "jest": "^27.2.0",
+    "jest-environment-jsdom-sixteen": "^2.0.0",
```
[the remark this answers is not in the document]
karlknowit: Nice catch! It did work by removing that dependency as well.

KenthLjungHBG:
A side-note about the commit messages: try to make them more informative about the commit itself. `fix: fix reviewcomments for getssmparameters function` loses all its meaning once we rebase this PR 🙂. A clearer message might be `fix: change getSsmParameters to named function`.

Also note that `fix: fix ...` is redudant, although I struggle with coming up with a good verb for fixes sometimes 😅.

services/booking-api/lambdas/cancel.js
```diff
@@ -0,0 +1,23 @@
...
+
+import * as response from '../../../libs/response';
+
+import { getSsmParameters } from '../helpers/getSsmParameters';
+import { sendBookingPostRequest } from '../helpers/sendBookingPostRequest';
+
+export const main = async event => {
+  const bookingId = event.pathParameters.id;
+
+  const [error, ssmParameters = {}] = await to(getSsmParameters());
```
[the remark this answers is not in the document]
karlknowit: You are correct, they should never be here. I will remove them

services/booking-api/lambdas/cancel.js
```diff
@@ -0,0 +1,23 @@
...
+  if (error) throwError(error.statusCode, error.message);
+
+  const { outlookBookingEndpoint, apiKey } = ssmParameters;
+  const url = `${outlookBookingEndpoint}/cancel`;
+  const body = { bookingId };
+
+  const [requestError] = await to(sendBookingPostRequest(url, apiKey, body));
+  if (requestError) throwError(requestError.status, requestError.errorMessage);
```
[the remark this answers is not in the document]
KenthLjungHBG: @dannil76 @TeddyAndersson Is this a good idea do you think? 🙂

dannil76 (changes requested):

services/booking-api/lambdas/cancel.js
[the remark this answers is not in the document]
karlknowit: From my point of view it is regarding having a function doing one thing and one thing only. I would like to have your feedback what we want to do in the project.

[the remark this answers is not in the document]
dannil76: Sounds reasonable.
Could take it even further and created an object with methods similar to: booking.create, booking.delete?
Have done something similar for vivaMs.

services/booking-api/README.md
```diff
@@ -0,0 +1,198 @@
...
+
+`GET`
+
+#### Endpoint
+
+`/booking/{bookingId}`
+
+#### JSON PAYLOAD
+
+`N/A`
```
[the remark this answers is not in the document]
karlknowit: Not necessary. I did remove the entire title.

services/booking-api/README.md
```diff
@@ -0,0 +1,198 @@
...
+```
+
+When you deploy the service, serverless will output the generated url in the terminal that the service can be accessed from.
+
+### Deploy and Run on AWS
+
+Deploy command:
+
+```bash
+$ sls deploy -v
```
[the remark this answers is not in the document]
karlknowit: Fixing

services/booking-api/babel.config.js
karlknowit: Yes, its because we are using ECM

services/booking-api/helpers/booking.js
```diff
@@ -0,0 +1,41 @@
+import * as request from '../../../libs/request';
+import params from '../../../libs/params';
+import config from '../../../config';
+
+const PATH = {
```
dannil76, line 5: Maybe we could name this URI_RESOURCE?
https://restfulapi.net/resource-naming/

services/booking-api/helpers/booking.js
```diff
@@ -0,0 +1,41 @@
...
+
+function cancel(body) {
+  return sendBookingPostRequest(PATH.CANCEL, body);
+}
+
+function get(body) {
+  return sendBookingPostRequest(PATH.GET, body);
+}
+
+async function sendBookingPostRequest(path, body) {
```
dannil76, line 27: The naming of this function implies that it will handle POST requests only. Can this be a problem in the future?

The 'get' function is also using this function. Is it correct to send post payload to the 'get' endpoint?
It's common to construct GET requests using path query parameters?

Sure I'm stupid not getting the picture :)

services/booking-api/lambdas/cancel.js
```diff
@@ -0,0 +1,18 @@
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
+import to from 'await-to-js';
+
+import * as response from '../../../libs/response';
+
+import booking from '../helpers/booking';
+
+export async function main(event) {
+  const bookingId = event.pathParameters.id;
+  const body = { bookingId };
```
dannil76, line 10: This looks somewhat confusing to me.
I would interpret booking Id to contain the actual id for the booking?

Curious what the 'body' means?

Maybe the pathParameter - id is lying in some regard?

services/booking-api/lambdas/cancel.js
```diff
@@ -0,0 +1,18 @@
...
+
+  const [error] = await to(booking.cancel(body));
+  if (error) {
+    throwError(error);
```
dannil76, line 14: throwError expects two args, statusCode and message. If error is the status code then this is fine I guess? :)

services/booking-api/lambdas/create.js
```diff
@@ -0,0 +1,18 @@
+import { throwError } from '@helsingborg-stad/npm-api-error-handling';
+import to from 'await-to-js';
+
+import * as response from '../../../libs/response';
+
+import booking from '../helpers/booking';
+
+export async function main(event) {
+  const body = { ...JSON.parse(event.body) };
```
dannil76, line 9: Curious about why you are spreading here?

services/booking-api/lambdas/create.js
```diff
@@ -0,0 +1,18 @@
...
+
+  const [error, createBookingResponse] = await to(booking.create(body));
+  if (error) {
+    throwError(error);
+  }
+
+  const { data } = createBookingResponse.data;
+  return response.success(200, data);
```
dannil76, line 17: We're trying to avoid using naming like 'data' because all thing is data and it is in some regards unclear what it is.
Maybe it could be named 'calendarBooking' or something in that direction? :)

services/booking-api/lambdas/update.js
```diff
@@ -0,0 +1,25 @@
...
+    throwError(cancelError);
+  }
+
+  body = { ...JSON.parse(event.body) };
+  const [createError, createBookingResponse] = await to(booking.create(body));
+  if (createError) {
+    throwError(createError);
+  }
+
+  const { data } = createBookingResponse.data;
```
dannil76, line 23: is data nested inside another data?

services/booking-api/serverless.yml
```diff
@@ -0,0 +1,77 @@
...
+functions:
+  createBooking:
+    handler: lambdas/create.main
+    events:
+      - http:
+          path: booking
+          method: post
+          private: true
+          cors: true
+  cancelBooking:
```
dannil76, line 54: Would like to have one line break after every lambda declaration. :)

dannil76 (changes requested):

services/booking-api/lambdas/update.js
karlknowit: Yes, thats correct

services/booking-api/lambdas/create.js
karlknowit: This is destructured from the Datatorget response. We are not naming it to data ourself.

services/booking-api/helpers/booking.js
karlknowit: Sounds like a good suggestion

libs/params.js
```diff
@@ -1,27 +1,29 @@
...
-
-export default params;
+import SSM from 'aws-sdk/clients/ssm';
+import to from 'await-to-js';
+
+const ssm = new SSM({ apiVersion: '2014-11-06' });
+
+const cachedParameters = {};
+
+async function read(name) {
```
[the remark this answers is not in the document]
karlknowit: I don't think that parameters should start with uppercase.

services/booking-api/serverless.yml
karlknowit: Me too, but I followed the syntax in other serverless.yaml files. But if thats OK with line breaks, then I will add it.

libs/params.js
```diff
@@ -1,27 +1,29 @@
...
-  read,
-};
-
-export default params;
+import SSM from 'aws-sdk/clients/ssm';
+import to from 'await-to-js';
+
+const ssm = new SSM({ apiVersion: '2014-11-06' });
+
+const cachedParameters = {};
```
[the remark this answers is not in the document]
karlknowit: Allright, will rename it

services/booking-api/lambdas/cancel.js
dannil76: My bad. Ignore

services/booking-api/lambdas/create.js
karlknowit: Shouldn't be needed. Nice catch. Will remove it.

services/booking-api/lambdas/create.js
TeddyAndersson: To ease the readability, i would suggest just naming the data object or the destructed data attribute to something mer descriptive. The reason is just to ease the understanding of what the data attribute contains.

My suggestion if it i'm understanding the code correct is that the data attribute is a booking object, then maybe an improvement could be
```JS
const { data: booking } = createBookingResponse.data
```

services/booking-api/serverless.yml
TeddyAndersson: It's OK, create the linebreak. And if you see this elsewhere when you are in the code base, it's fine to correct (y)

services/booking-api/lambdas/create.js
TeddyAndersson: The spreading is due to that the event.body is a JSON payload so that's fine. But i wounder what the JSON data represents ie what does the body contains, is it a new booking object?
Then i would suggest something like the following:
```JS
const bookingEventtBody = { ...JSON.parse(event.body) }
```

libs/params.js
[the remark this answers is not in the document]
TeddyAndersson: Have too agree with @karlknowit here, since we use camelCase we should stick to that syntax.

libs/params.js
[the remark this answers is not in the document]
TeddyAndersson: This is somewhat a new insight @karlknowit, so you will probably find code that break this standard.
But the reason for postfixing with List and Collection when needed is to ease the readability.

```JS
// easy to distinguish 
userList = [...]
user = {}

// harder to distinguish
users = []
user = {}
```

Just so you know the reasoning we have had around this :)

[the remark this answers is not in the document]
TeddyAndersson: Might be a poor example, but i think you get the point!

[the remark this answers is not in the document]
TeddyAndersson: it's even harder when you have something like userFirstnames and userFirstname as an example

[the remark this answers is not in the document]
karlknowit: Normally you distinguish this kinds of stuff with the suffix **s** as you mention. But I guess its OK to add the word list in it as well.

services/booking-api/lambdas/cancel.js
karlknowit: Not entirely correct. throwError expects 2 arguments but the second one is optional since it will receive a default value if no message is added.

TeddyAndersson (approved):
Happy with these changes, feel free to rebase (y)

services/booking-api/lambdas/update.js
```diff
@@ -0,0 +1,23 @@
...
+  const [cancelError] = await to(booking.cancel(bookingId));
+  if (cancelError) {
+    return response.failure(cancelError);
+  }
+
+  const body = JSON.parse(event.body);
+  const [createError, createBookingResponse] = await to(booking.create(body));
+  if (createError) {
+    return response.failure(createError);
+  }
```
KenthLjungHBG, line 19: Maybe off-topic but how do we handle if `cancel` succeeds but `create` fails? Can we rollback `cancel`, or must the API-user now use `create` instead of `update`?

This is probably a larger architectural question and doesn't need any code considerations right now 🙂.

KenthLjungHBG (approved):
Looks nice!

services/booking-api/lambdas/update.js
karlknowit: This is a good question. I think I will create a task about this and add it to the backlog. Because I think this is something that we need to follow up.

TeddyAndersson: I guess this also depends on how data-torget is structured. So it's not fully on our side of the pitch so to speak. But if we feel a need for an update in the future, we can add it then i guess (y)

@karlknowit probably have more insight into how data-torget works and if it makes sense to have an update (PUT or POST) request in our apis facing the API-user :)

karlknowit: I think we have a need for update, and to update an outlook booking we first need to cancel the first one and thereafter creating a new one. I know that this isn't the most ideal solution but thats how its implemented in Datatorget.

TeddyAndersson:
And again sorry for a messy review process for this PR, and i get that it's not always clear to find the correct code standard since it has evolved over time and thereby every lambda does not follow the standard. So it's fine if you are confused/frustrated. We have a technical debt to tackle in order to fix these issues regarding code standards :)

TeddyAndersson:
> And again sorry for a messy review process for this PR, and i get that it's not always clear to find the correct code standard since it has evolved over time and thereby every lambda does not follow the standard. So it's fine if you are confused/frustrated. We have a technical debt to tackle in order to fix these issues regarding code standards :)

Also want to mention that most of the insight regarding code standards comes from Clean Code, so if you some time over i would recommend you to read it :D

PR #229, opened 2021-09-30
Feat/search api service

sepehrnoor:
## What was solved

Added a new service (search-api) for making requests against Datatorget to handle outlook calendar search request. More search-related endpoints (searching time slots for municipal workers, searching users, etc.) can be added to the API as needed in the future.

## How was it solved

It was solved by creating a new serverless stack containing a lambda function which makes queries against Datatorget, as well as an API gateway resource for accessing it via HTTP.

This structure tries to follow the agreed way of creating serverless stacks in the rest of the repository.

## How was it tested

It was tested by deploying the stack to my sandbox and making successful requests against the API Gateway path from Postman, as well as from within the AWS UI.


services/search-api/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "author": "Helsingborg Stad",
+  "license": "MIT",
+  "dependencies": {},
+  "devDependencies": {
+    "@babel/core": "^7.15.5",
+    "@babel/plugin-transform-modules-commonjs": "^7.15.4",
+    "babel-jest": "^27.2.0",
+    "jest": "^27.2.0",
+    "jest-environment-jsdom-sixteen": "^2.0.0"
+  }
```
dannil76, line 19: Maybe this should live in the package.json at root-level? We try to put global packages in here for other services to use.

dannil76 (changes requested):
Some minor styling :)

services/search-api/package.json
karlknowit: Yes, these dependencies are actually not needed in the service itself. They do exists in the root package.json

dannil76: Even better :)

jest.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   transform: {
     '^.+\\.[t|j]sx?$': 'babel-jest',
   },
+  testEnvironment: 'jest-environment-jsdom-sixteen',
```
karlknowit, line 5: This isn't needed

services/search-api/helpers/search.js
```diff
@@ -0,0 +1,38 @@
...
+import to from 'await-to-js';
+
+import * as request from '../../../libs/request';
+import params from '../../../libs/params';
+import config from '../../../config';
+
+async function searchBookings(body) {
+  const { outlookSearchEndpoint, apiKey } = await getSsmParameters();
+
+  if (!outlookSearchEndpoint || !apiKey) {
```
karlknowit, line 11: This check is unnecessary since thats already done in params.read function

karlknowit (changes requested):

services/search-api/lambdas/searchBookings.js
```diff
@@ -0,0 +1,18 @@
...
+import * as response from '../../../libs/response';
+
+import { searchBookings } from '../helpers/search';
+
+export async function main(event) {
+  const body = { ...JSON.parse(event.body) };
+
+  const [requestError, getSearchResponse] = await to(searchBookings(body));
+  if (requestError) {
+    throwError(requestError.status, requestError.errorMessage);
```
karlknowit, line 13: I think that the lambda should return an error response instead of throwing. Check **response.failure** function on other lambdas

karlknowit (changes requested):

services/search-api/helpers/search.js
```diff
@@ -0,0 +1,38 @@
...
+    await getSsmParameters();
+  }
+
+  const requestClient = request.requestClient(
+    { rejectUnauthorized: false },
+    { 'X-ApiKey': apiKey }
+  );
+
+  const url = `${outlookSearchEndpoint}`;
```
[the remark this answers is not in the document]
sepehrnoor: Good catch! Fixed.

services/search-api/package.json
sepehrnoor: Removed dependencies and ran the tests again, everything works fine.

dannil76 (approved):

karlknowit (approved):

PR #234, opened 2021-10-12
Feat/dynamic pdf generator

TeddyAndersson:
## What was solved
Generating pdfs from html files stored in s3.

## How was it solved
This was solved by using a headless chromium browser and use the html to pdf functionality that chromium have built in. The headless chromium browser is called puppeteer requires a heavy package so in order to get this to work a layer with the required packages was needed in order to run puppeteer.

The pdf-generate lambda is a simple implementation that **(1)** reads an html file from the pdf-storage s3 bucket, **(2)** starts up a puppeteer headless browser **(3)** uses the browser to generate the pdf from the retrieved html-file and then **(4)** saves the pdf to the pdf-storage s3 bucket. **(4)** After successful generating a pdf and storing it an event is passed to the AWS event-bridge.

This solution have been added due to that generating pdf from html is a tested solution. This solution also comes with another benefit which is no need to implement a custom way of transforming a data-structure to pdf. Html in itself is also simple to customise and work with as a developer in styling, layout etc.

## How was it tested
1. Create a simple html file in and upload it to the pdf-storage s3 bucket.
2. Create an event where the html-file's s3 object-key is passed.
3. Check to see if a pdf is stored in the pdf-storage s3 bucket.

### DIFF
--- a/services/html-pdf-ms/serverless.yml
+++ b/services/html-pdf-ms/serverless.yml
@@ -1,4 +1,4 @@
-service: pdf-ms
+service: html-pdf-ms
 
 plugins:
   - serverless-bundle


helsingborgsstad:
Task linked: [CU-webc6d Skapa logik som genererar pdf från renderad html](https://app.clickup.com/t/webc6d)

PR #243, opened 2021-11-01
feat: add parameter store implementation of bookables service

sepehrnoor:
## What was solved

Added a new service (bookables-api) for listing bookable services for Mitt Helsingborg. The data is retrieved from the parameter store for now. Depends on https://github.com/helsingborg-stad/helsingborg-io-sls-resources/pull/75 being approved to avoid pipeline trouble.

## How was it solved

It was solved by creating a new API that retrieves the relevant information from the parameter store.

Optimally we would create a platform with a UI and perhaps a dynamo for storing the bookable services, but due to time constraints (and the relative immutability of the bookable services offered by the municipality) we opted for a simple UI-less parameter store solution for now.
Even though the code structure might seem somewhat redundant, it was written to be consistent with the rest of the repository and open for further development (perhaps to the aforementioned platform).

## How was it tested

It was tested by deploying the stack to my sandbox and making successful requests against API Gateway from the AWS UI, as well as from within the application in a demo.

### DIFF
--- a/services/bookables-api/README.md
+++ b/services/bookables-api/README.md
@@ -1,0 +1,99 @@
+# HELSINGBORG IO SLS BOOKABLES SERVICE
+
+## Purpose
+
+The Bookables Service's purpose is to provide the app with data related to services that can be booked at Helsingborg Stad.
+
+## Description
+
+The Bookables Service is an API that returns information about the bookable services at Helsingborg Stad. Currently it retrieves the data from the parameter store, for ease of deployment. In the future this could be developed further into a platform with its own dynamo.
+
+## Getting started
+
+1. Read the global requierments for this repo, can be found [here](https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/dev/README.md)
+
+### AWS API GATEWAY
+
+A running instance of an API GATEWAY on AWS that includes a gateway resource named /search. You can find and deploy this in our [resource](https://github.com/helsingborg-stad/helsingborg-io-sls-resources/tree/dev/services/gateway/resources/search) repository.
+
+### AWS PARAMETERSTORE (OPTIONAL)
+
+A setup of search AWS paramterstore on aws. This can be created from the resource api. You can find and deploy this in our [resource](https://github.com/helsingborg-stad/helsingborg-io-sls-resources/tree/dev/services/parameterStore) repository.
+
+### Installation
+
+```bash
+$ npm install
+```
+
+### Run Local
+
+```bash
+$ sls offline
+```
+
+When you deploy the service, serverless will output the generated url in the terminal that the service can be accessed from.
+
+### Deploy and Run on AWS
+
+Deploy command:
+
+```bash
+$ sls deploy -v
+```
+
+When you deploy the service, serverless will output the generated url in the terminal that the service can be accessed from.
+
+### Tear down service
+
+Teardown command:
+
+```bash
+$ sls remove
+```
+
+When you tear down the service, serverless will remove all resources created in the AWS console, including emptying the `deploymentBucket`.
+
+### Testing (Jest)
+
+Run test command:
+
+```bash
+$ npm run test
+```
+
+Run test command with [options](https://jestjs.io/docs/cli#options):
+
+```bash
+$ npm run test -- --[option_1] --[option_2]
+```
+
+## API
+
+### GET BOOKABLES
+
+#### Request Type
+
+`GET`
+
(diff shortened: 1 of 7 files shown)


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/243)

services/bookables-api/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+The Bookables Service's purpose is to provide the app with data related to services that can be booked at Helsingborg Stad.
+
+## Description
+
+The Bookables Service is an API that returns information about the bookable services at Helsingborg Stad. Currently it retrieves the data from the parameter store, for ease of deployment. In the future this could be developed further into a platform with its own dynamo.
+
+## Getting started
+
+1. Read the global requierments for this repo, can be found [here](https://github.com/helsingborg-stad/helsingborg-io-sls-api/blob/dev/README.md)
```
karlknowit, line 13: requierments -> requirements

services/bookables-api/README.md
```diff
@@ -0,0 +1,99 @@
+# HELSINGBORG IO SLS BOOKABLES SERVICE
+
+## Purpose
+
+The Bookables Service's purpose is to provide the app with data related to services that can be booked at Helsingborg Stad.
+
+## Description
+
+The Bookables Service is an API that returns information about the bookable services at Helsingborg Stad. Currently it retrieves the data from the parameter store, for ease of deployment. In the future this could be developed further into a platform with its own dynamo.
```
karlknowit, line 9: I don't think that we should write what the service could be. Instead only describe how the service is at this particular moment.

services/bookables-api/README.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+When you deploy the service, serverless will output the generated url in the terminal that the service can be accessed from.
+
+### Deploy and Run on AWS
+
+Deploy command:
+
+```bash
+$ sls deploy -v
```
karlknowit, line 42: I think that we can remove the -v command. sls deploy is enough

services/bookables-api/README.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+When you deploy the service, serverless will output the generated url in the terminal that the service can be accessed from.
```
karlknowit, line 45: The sentence:
When you deploy the service, serverless will output the generated url in the terminal that the service can be accessed from

appears twice, is it meant to be like that?

services/bookables-api/helpers/bookables.js
```diff
@@ -0,0 +1,21 @@
...
+
+import params from '../../../libs/params';
+import config from '../../../config';
+
+async function getBookables() {
+  const { bookables } = await getSsmParameters();
+  return bookables;
+}
+
+async function getSsmParameters() {
```
karlknowit, line 12: I am having an idea that we can remove this function `getSsmParamters` and instead use the `throwError` function directly in the params.read function. I think that would be better instead of duplicate this function all the time.

services/bookables-api/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "bookables-api",
+  "version": "0.0.1",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "node ../../node_modules/.bin/jest"
+  },
+  "keywords": [],
```
karlknowit, line 9: Keywords can be completely removed.

karlknowit (changes requested):

services/bookables-api/helpers/bookables.js
sepehrnoor: Yes, we should discuss this with the rest of the team and do a separate pull request for it where we remove the function from all services. 
For now we'll keep the function for the sake of consistency with the other services.

services/bookables-api/README.md
sepehrnoor: Good idea, I have updated the README.

services/bookables-api/README.md
sepehrnoor: Both the offline and online deploy return an URL for accessing, so this sentence is repeated for both. 
But I agree, it's unnecessary repetition. Removed.

karlknowit (approved):
Looks good!

sepehrnoor:
Merging because this service does not interact with or affect anything else.

PR #255, opened 2021-11-09
Feat/office auth

karlknowit:
# What was solved
Added a lambda authorizer that is going to be used for authenticating Office users when they are using our add-ins. The lambda authorizer is used to prevent users not authenticated with Azure AD to access our AWS API:s for datatorget.

# How was it solved
Solved it by adding a new lambda authorizer called OfficeAuthorizer. This lambda functions decodes and verifies the accessToken that is going to be sent from the add-ins when they are doing requests against our API:s. If it's not a valid token, then the request is then unauthorised.

We didn't want to create a separate authentication flow for an Office user, and since they were already authenticated against Azure AD, we wanted to utilize the credentials received from it.

# How was it tested
Tested it by creating a dummy lambda function which was protected by the new lambda authorizer. Created an access token from the Office API and sent a request from postman with the token in the authorization header.

Also did try the authorizer with a "broken" access token and which gave me an unauthorised response.


sebastianthulin:
Quality gate Quality gate: **_Failed_** 
 


            
 #### 6 Issues 
* Code issue 6 Code issues 

 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/255)

helsingborgsstad:
Task linked: [CU-1jnu9z4 Skapa lambda authorizer för Office användare](https://app.clickup.com/t/1jnu9z4)

karlknowit:
This PR is a replacement for the old one:
https://github.com/helsingborg-stad/helsingborg-io-sls-api/pull/250

Which has already been approved by @sepehrnoor and @dannil76. 
I will therefore take the opportunity to merge this PR.

PR #262, opened 2021-11-15
feat: added get endpoint for fetching historical attendees

karlknowit:
# What was solved
Added new endpoint for fetching historical attendees, i.e. attendees which is used as a users "network" within the application.

# How was it solved
Added a new GET endpoint which takes the referenceCode as path parameter and startTime and endTime as required query string parameters.

The reason for using GET rather than POST method is because we want the endpoint to be as much REST as possible. 

# How was it tested
Tested created a new calendar booking with a specific referenceCode. Took that referenceCode and tried it out with the new endpoint. Checked the result that was returned in postman.

### DIFF
--- a/services/booking-api/helpers/booking.js
+++ b/services/booking-api/helpers/booking.js
@@ -10,6 +10,7 @@ const URI_RESOURCE = {
   CANCEL: 'booking/cancel',
   GET: 'booking/get',
   SEARCH: 'booking/search',
+  GET_HISTORICAL_ATTENDEES: 'booking/getHistoricalAttendees',
 };
 
 const METHOD = {
@@ -30,6 +31,10 @@ function get(bookingId) {
 
 function search(body) {
   return sendBookingPostRequest(URI_RESOURCE.SEARCH, body);
+}
+
+function getHistoricalAttendees(body) {
+  return sendBookingPostRequest(URI_RESOURCE.GET_HISTORICAL_ATTENDEES, body);
 }
 
 async function sendBookingPostRequest(path, body) {
@@ -59,4 +64,4 @@ async function getSsmParameters() {
   return response;
 }
 
-export default { create, cancel, get, search };
+export default { create, cancel, get, search, getHistoricalAttendees };
--- a/services/booking-api/lambdas/getHistoricalAttendees.js
+++ b/services/booking-api/lambdas/getHistoricalAttendees.js
@@ -5,8 +5,6 @@ import * as response from '../../../libs/response';
 import booking from '../helpers/booking';
 
 export async function main(event) {
-  console.log('EVENT: ', JSON.stringify(event, undefined));
-
   let referenceCode = event.pathParameters?.referenceCode;
 
   if (!referenceCode) {
--- a/services/booking-api/lambdas/getHistoricalAttendees.js
+++ b/services/booking-api/lambdas/getHistoricalAttendees.js
@@ -1,0 +1,39 @@
+import to from 'await-to-js';
+
+import * as response from '../../../libs/response';
+
+import booking from '../helpers/booking';
+
+export async function main(event) {
+  console.log('EVENT: ', JSON.stringify(event, undefined));
+
+  let referenceCode = event.pathParameters?.referenceCode;
+
+  if (!referenceCode) {
+    return response.failure({
+      status: 403,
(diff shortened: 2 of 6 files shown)


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/262)

helsingborgsstad:
Task linked: [CU-1fmc933 Bygg ny endpoint för att ta fram nätverk för användare](https://app.clickup.com/t/1fmc933)

services/booking-api/lambdas/getHistoricalAttendees.js
```diff
@@ -0,0 +1,39 @@
...
+    return response.failure({
+      status: 403,
+      message: 'Missing required parameter: "referenceCode"',
+    });
+  }
+
+  referenceCode = decodeURIComponent(referenceCode);
+  const { startTime = '', endTime = '' } = event.queryStringParameters;
+
+  if (!startTime | !endTime) {
```
dannil76, line 22: Have a somewhat hard time to understand this expression. Maybe it could be extraced to an function that's describes the check in a clear way?

services/booking-api/lambdas/getHistoricalAttendees.js
```diff
@@ -0,0 +1,39 @@
...
+import * as response from '../../../libs/response';
+
+import booking from '../helpers/booking';
+
+export async function main(event) {
+  console.log('EVENT: ', JSON.stringify(event, undefined));
+
+  let referenceCode = event.pathParameters?.referenceCode;
+
+  if (!referenceCode) {
```
dannil76, line 12: This kind of neg-checking is a bit hard to spot. Maybe the check could be a bit more verbose?

services/booking-api/lambdas/getHistoricalAttendees.js
```diff
@@ -0,0 +1,39 @@
+import to from 'await-to-js';
+
+import * as response from '../../../libs/response';
+
+import booking from '../helpers/booking';
+
+export async function main(event) {
+  console.log('EVENT: ', JSON.stringify(event, undefined));
```
dannil76, line 8: You could use the logger helper function from libs?

services/booking-api/helpers/booking.js
```diff
@@ -10,6 +10,7 @@ const URI_RESOURCE = {
   CANCEL: 'booking/cancel',
   GET: 'booking/get',
   SEARCH: 'booking/search',
+  GET_HISTORICAL_ATTENDEES: 'booking/getHistoricalAttendees',
 };
```
dannil76, line 14: Easy to missunderstand this general constant if you think CANCEL is eq to the string 'CANCEL'?
Maybe it could be named more precise like CANCEL_BOOKING = '...' etc

dannil76 (changes requested):

services/booking-api/lambdas/getHistoricalAttendees.js
karlknowit: The neg-checking is actually one of the better methods to use if you want to check for falsy values like undefined, false, null and empty string at the same time. Otherwise you could end up with multiple checks like:

```js
if (referenceCode === undefined | referenceCode === '' | referenceCode === false. ...) { ... }
```

Another solution would be to default referenceCode to a falsy value like:

```js
let referenceCode = event.pathParameters?.referenceCode || '';

if (referenceCode === '') { ... }
```

services/booking-api/lambdas/getHistoricalAttendees.js
karlknowit: Will remove the console.log and create a subtask where we add logging of events in all booking-api lambdas in the same way.

services/booking-api/helpers/booking.js
karlknowit: Do you think so even when the constant name is URI_RESOURCE?

services/booking-api/lambdas/getHistoricalAttendees.js
karlknowit: Lets have a discussion about this one

dannil76 (approved):

TeddyAndersson (approved):
Ship it!

PR #273, opened 2021-11-22
Fix/pass state in event

TeddyAndersson:
I've added the possibility of reacting to the case state after a case have been updated through the caseUpdate lambda. The generateRecurringCaseHtml lambda listen and reacts to this new detail attribute.

The generateRecurringCaseHtml fired on occasions when it should not. This is due to that the status ("active:submitted") is set on several occasions throughout a case's life-span, but the pdf should only be generated once when state is "VIVA_CASE_CREATED" and the status is "active:submitted".

### DIFF
--- a/services/cases-api/lambdas/updateCase.js
+++ b/services/cases-api/lambdas/updateCase.js
@@ -173,7 +173,7 @@ export async function main(event, context) {
 
   const [putEventError] = await to(
     putEvent(
-      { caseKeys, status: newCaseStatus },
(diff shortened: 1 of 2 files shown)


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/273)

dannil76 (approved):
Looks right. Testing it now

PR #275, opened 2021-11-23
Fix/spacing beetween tables

TeddyAndersson:
Added margin at the bottom of each table in the handlebars template, in order for tables to not appear as one in the generated pdf.


dannil76 (approved):
Hard to test

PR #281, opened 2021-12-02
Feat/build template notes data from tags

TeddyAndersson:
## Feature description
Added support for parsing multiple notes answers into a handlebars template.

## Solution description
Before the creation of notes in the template helper only supported the retrieval of one field based on a hardcoded id. 

Now the creation of notes in the template helper is based on retrieval of multiple notes answers based on tags. We simply filter out answers that contains the tags `notes`, `note` and then a note object is created for each answer and return as an array of notes to the template.

viva-ms 

## Is there any existing behavior change of other features due to this code change?
Yes. The form needs to be updated with correct tagging on certain fields in order for notes to be passed into the handlebars template. If this is not corrected the pdf that is generated in the service html-pdf-ms will not contain any notes at all. 

No

### DIFF
--- a/.deploy/cloudformation/pipelines/release/stack.yml
+++ b/.deploy/cloudformation/pipelines/release/stack.yml
@@ -48,8 +48,6 @@ Resources:
         Location: !Sub "${S3BucketArtifact}/cache"
         Type: S3
 
-
-
   ##
   # CodePipeline
   ##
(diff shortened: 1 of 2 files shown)


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/281)

dannil76 (approved):

PR #282, opened 2021-12-10
Feat/sync viva children

dannil76:
## Feature description
Added functionality for the case to be pre-filled with information about children from the case in Viva, before the applicant begins the application in the app.

## Solution description
Created help functions based on a previous solution of filling in the case in advance. vivaCreateCase has been expanded to check if there are children on the case from Viva and send in the list to a function together with previous answers from the 'ongoing' application together with the template for the 'ongoing' application.

For each child in the list, we pick out the attributes firstName, lastName and personalNumber and add these to an answer list, where we find form fields with the tags 'children'

viva-ms lambda createVivaCase together with helper function populateFrom

Recurring form needs to be updatead with an repeaterFiled with correct id

Below you see what a repaterFiled should look like for children in the form 'recurring'
`{
  "description": "Lägg till barn",
  "label": "",
  "id": "childrenInfo",
  "type": "repeaterField"
  "addButtonText": "Lägg till barn",
  "inputSelectValue": "repeaterField",
  "heading": "Uppgifter om barn",
  "inputs": [
    {
      "id": "childrenFirstname",
      "type": "text",
      "title": "Förnamn",
      "inputSelectValue": "text",
      "validation": {
        "isRequired": true,
        "rules": [
          {
            "method": "isEmpty",
            "message": "Du får inte lämna detta fält tomt",
            "validWhen": false
          }
        ]
      },
      "tags": [
        "children",
        "firstName",
        "group:children:info:x"
      ]
    },
    {
      "id": "childrenLastname",
      "title": "Efternamn",
      "type": "text",
      "inputSelectValue": "text",
      "validation": {
        "isRequired": true,
        "rules": [
          {
            "method": "isEmpty",
            "message": "Du får inte lämna detta fält tomt",
            "validWhen": false
          }
        ]
      },
      "tags": [
        "children",
        "lastName",
        "group:children:info:x"
      ]
    },
    {
      "id": "childrenPersonalID",
      "title": "Personnummer",
      "type": "number",
      "inputSelectValue": "number",
      "validation": {
        "isRequired": true,
        "rules": [
          {
            "args": {
              "options": {
                "no_symbols": true
              }
            },
            "validWhen": true,
            "method": "isNumeric",
            "message": "Du måste ange en siffra"
          },
          {
            "method": "isEmpty",
            "message": "Du får inte lämna detta fält tomt",
            "validWhen": false
          }
        ]
      },
      "tags": [
        "children",
        "personalNumber",
        "group:children:info:x"
      ]
    }
  ]
}
`

## Covered unit tests cases?
Yes, tests are in the test folder in the service project root
run
`cd viva-ms && yarn`
followed by
`yarn test`


sebastianthulin:
Quality gate Quality gate: **_Failed_** 
 


            
 #### 4 Issues 
* Code issue 4 Code issues 

 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/282)

helsingborgsstad:
Task linked: [CU-1n5v3r7 Populera svar för barn repeater-fält med person-data från ärendet](https://app.clickup.com/t/1n5v3r7)

services/viva-ms/helpers/populateForm.js
```diff
@@ -0,0 +1,34 @@
+import { generateDataMap, formatAnswer, mergeAnswers } from '../../../libs/formAnswers';
+
+export function populateChildrenAnswers(repeaterInputList, childrenList) {
+  const answers = [];
+
+  childrenList.forEach((child, childrenListIndex) => {
```
TeddyAndersson, line 6: This conversion from childrenList to answerList could be done in a more modern fashion by using reduce/map.
What we want is to create answers from children and repeaterInput values. This is a perfect scenario for the usage of the reduce function.

```JS
const children = [{
  role: 'child',
  personalNumber: 1990123456,
  firstName: 'John',
  lastName: 'Doe',
}, ...]
const answers = children.reduce((answers, child, childIndex) => {
  
  const childAnswers = Object(child).keys.map(key => {
     // create child answer from inputs
     const answer = repeaterInputList.map((input) => {
         if (input.tags.includes(key){
           const inputId = input.id.replace('[*]', childIndex)
           return formatAnswer(inputId, input.tags, child[key])
         }
       }
     )  
  })
  
  return [...answers, ...childAnswers]
}, [])

```

services/viva-ms/helpers/populateForm.js
```diff
@@ -0,0 +1,34 @@
...
+        .filter(Boolean);
+
+      answers.push(...childAnswer);
+    }
+  });
+
+  return answers;
+}
+
+export default function populateFormWithVivaChildren(form, formTemplate, vivaChildrenList) {
```
TeddyAndersson, line 24: This function is very redundant, in terms of what the  populateFormWithPreviousCaseAnswers already does. Is there a way to combine these two functions without some effort. 

I'm aware that the populate code in general is very messy, but making it messier is not a good practice. Better to refactor as we go (y)

TeddyAndersson (commented):
Left some thoughts and suggestions in order to improve the code (y)

libs/formAnswers.js
```diff
@@ -1,4 +1,4 @@
-function generateDataMap(form) {
+export function generateDataMap(form) {
```
TeddyAndersson, line 1: Suggestion here is to rename this function to something along the lines with getFormFieldsToPopulate() or getFormFieldsWithLoadPreviousAttribute()

services/viva-ms/helpers/populateForm.js
dannil76: Refactored. Not sure if it's less readable :)

services/viva-ms/helpers/populateForm.js
dannil76: They do different things. Will keep for now.

libs/formAnswers.js
dannil76: Will keep for now

TeddyAndersson:
I'm happy with the changes that have been made. However I think it would be a good idea to track the technical debt that we are creating here in terms of the comments that are not addressed. But then again these kind of tasks have a way of not being addressed or forgotten, so strongly recommend that we just fixit now instead of later :)

TeddyAndersson (approved):

libs/formAnswers.js
TeddyAndersson: Is this not a quick fix? That would improve readability? And when will these changes get addressed otherwise?

services/viva-ms/helpers/populateForm.js
TeddyAndersson: The code tells a different story, they almost look identical. Except from the handling of childrenList :)
But if it's to much of an effort i will let this of the hook :)

libs/formAnswers.js
dannil76: Done!

PR #283, opened 2021-12-13
fix: make gethistoricalattendees not timeout

karlknowit:
# What was solved
Solved the issue were getHistoricalAttendees request would fail because of axios timeout.

# How was it solved
Solved it by increasing the requestTimeout when when doing the getHistoricalAttendees request by adding an environment variable to the lambda function. 

Since the request takes more time than 5 seconds, we wanted to increase it in order for the request to be successful. By adding the timeout as an environment variable, we can manually change it for troubleshooting if it happens again.

# How was it tested
Tested it by deploying the solution and see if the request did timeout or not.

### DIFF
--- a/services/booking-api/helpers/booking.js
+++ b/services/booking-api/helpers/booking.js
@@ -34,15 +34,17 @@ function search(body) {
 }
 
 function getHistoricalAttendees(body) {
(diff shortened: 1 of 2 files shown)


sepehrnoor (approved):
Shippit!

PR #286, opened 2021-12-13
fix: check if timeslot is available before creating booking

karlknowit:
# What was solved
Check if timeslot is available before creating a booking

# How was it solved
Solved it by first search within startTime and endTime if a booking is returned or not. If the response come back empty, then we can create a new booking. Otherwise return failure.

Searching for a booking with datatorget search endpoint is the only way to check if we can create a new booking within given timespan or not. 

# How was it tested
Tested it by deploying the solution, created a booking within a timespan and tried to create a new booking with the same timespan. The second attempt failed, which means that the logic works.

### DIFF
--- a/services/booking-api/lambdas/create.js
+++ b/services/booking-api/lambdas/create.js
@@ -17,6 +17,17 @@ export async function main(event) {
     });
   }
 
+  const searchBookingBody = { startTime, endTime };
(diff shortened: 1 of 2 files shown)


sepehrnoor (approved):
Looks good to me!

PR #287, opened 2021-12-14
feat: single deployment bucket

petter-a:


petter-a:
Replaced with a more scoped PR that only adds the deployment bucket and not any further refactoring

PR #304, opened 2022-01-18
Feat/update linting tools

petter-a:
Updates eslint configuration to reflect common rules and support typescript.
Also a separate prettier config is introduced.


sebastianthulin:
 


            
 [Details](https://app.embold.io/organization/gh/helsingborg-stad/repositories/602c13c73b93a7537d767e421a43afaa/pull-request/304)

helsingborgsstad:
Task linked: [CU-1pe1ex2 Uppdatera .eslintrc med nyare config (och ändra till .eslintrc.js)](https://app.clickup.com/t/1pe1ex2)

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+    semi: true,
+    singleQuote: true,
+    tabWidth: 4,
```
dannil76, line 4: What is the reason to change from 2 spaces to 4?
I vote for keeping it as is ie. 2 spaces

.eslintrc.js
```diff
@@ -0,0 +1,95 @@
...
+                'linebreak-style': ['error', 'unix'],
+                'callback-return': ['error'],
+                'block-spacing': ['error'],
+                'comma-style': ['error', 'last'],
+                'eol-last': ['error', 'always'],
+                'lines-between-class-members': ['error', 'always'],
+                'max-len': [
+                    'error',
+                    {
+                        code: 120,
```
dannil76, line 76: What is the reason for this value? 120 chars is way to long. My suggestion is to keep it between 70 - 80 chars

dannil76 (changes requested):
Left some comments.

This branch should only deal with rules and settings for linting, etc. Believes that we developers should talk to each other about how we should handle the transition to applying these lint changes.

.prettierrc.js
petter-a: Guess it is a matter of preference. I think code is way more readable that way. Have reverted it to 2 though.

.eslintrc.js
petter-a: Again a preference thingy. Typescript prints better on a wider width. I have reverted to 80.

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  semi: true,
+  singleQuote: true,
+  tabWidth: 2,
+  printWidth: 80,
+};
```
MasterKenth, line 6: What is the motivation for these values? Ideally in my opinion this should be left blank (`{}`) as the prettier defaults should be all we need.

.eslintrc.js
```diff
@@ -0,0 +1,95 @@
...
+        'no-useless-catch': ['error'],
+        'no-useless-return': ['error'],
+        'no-empty': ['error'],
+        'no-extra-boolean-cast': ['error'],
+        'no-floating-decimal': ['error'],
+        'no-warning-comments': ['error'],
+        'no-irregular-whitespace': ['error'],
+        'no-delete-var': ['error'],
+        'no-undef-init': ['error'],
+        'no-redeclare': ['error'],
```
MasterKenth, line 55: Why are all these rules explicitly defined? The ideal eslint config in my opinion should be very sparse. Take a look at the [app config](https://github.com/helsingborg-stad/app-mitt-helsingborg/blob/develop/.eslintrc.js) for example.

MasterKenth (changes requested):
It's not clear what the motivation for exactly these configurations come from.

I also don't think this PR should include formatting all files according to the new rules. That should ideally be in a separate PR or done on a file-by-file basis if that is too conflicting. Is formatting automatically applied?

PR #309, opened 2022-01-20
feat: add lambda for checking the status of a users application version

karlknowit:
Added a new api (version-api) for fetching a status if a users application version of Mitt Helsingborg needs to be updated, can be updated or if update is required.

To make sure that a user of Mitt Helsingborg is using an application version that our backend supports, we want to check that the version used is within `min` and `max` version numbers configured in our backend.

If a version is below `min` we want to force the user to update the application.
If a version is between `min` and `max` we (possibly) want to recommend the user to update.
If a version is the same as `max`, then everything is OK.

## Explain your solution
The lambda function fetches the configurations from the SSM parameter store and (which is manually configured) and checks the `User-Agent` header sent from the application. From the `User-Agent` header, the application retrieves the device OS and application version used by that device and calculates the status depending on the SSM paramters.

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for service `status-api`
3. Make sure you have `version` gateway path deployed in resources repo
4. Make sure that you have SSM Parameter store values deployed which follows the format in this clickup: https://app.clickup.com/t/1few6ak
5. Run the endpoint from postman or application

## Tested environments


helsingborgsstad:
Task linked: [CU-1dftv05 Skapa API-endpoint "/version" för att hämta versionsdetaljer](https://app.clickup.com/t/1dftv05)

services/version-api/helpers/decideVersionStatus.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Decides the status to return depending on currentVersion
+ *
+ * @param {string} currentVersion The current application version (in semver format)
+ * @param {string} minVersion The minimum allowed application version (in semver format)
+ * @param {string} maxVersion The latest application version released (in semver format)
+ * @returns {string} Version status
+ */
+function decideVersionStatus(currentVersion, minVersion, maxVersion) {
```
dannil76, line 19: We have previously agreed that comments on functions should not be written. The function names should have as explanatory name as possible.

It is good to try to keep the number of arguments to as few as possible as this facilitates testing of the function. Preferably 1 but 2 is ok

services/version-api/helpers/getVersionConfigurations.js
```diff
@@ -0,0 +1,21 @@
...
+import to from 'await-to-js';
+
+import params from '../../../libs/params';
+import config from '../../../config';
+
+/**
+ * Fetch application version configurations from AWS SSM Parameter store
+ *
+ * @returns {Promise<{versions: {ios: {min: string, max: string}, android: {min: string, max: string}}}>} SSM parameters
+ */
```
dannil76, line 11: See previous comments :)

services/version-api/helpers/getVersionConfigurations.js
```diff
@@ -0,0 +1,21 @@
...
+async function getVersionConfigurations() {
+  const [error, response] = await to(params.read(config.version.envsKeyName));
+  if (error) {
+    throwError(500);
```
dannil76, line 15: This helper (throwError) should only be used in the main lambda function. Instead you could use the javascript 'throw' together with an relevant error message for this context.

services/version-api/lambdas/getVersionStatus.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Function that decides if a users application version of Mitt Helsingborg
+ * needs to be updated, can be updated or if version used is ok.
+ *
+ * @param {object} event The lambda event
+ * @param {object} event.headers
+ * @param {string} event.headers.User_Agent
+ * @returns {Promise<string>} The version status (i.e "OK", "UPDATE_OPTIONAL" or "UPDATE_REQUIRED")
+ */
```
dannil76, line 20: See comments on keep comments of this kind out

services/version-api/lambdas/getVersionStatus.js
```diff
@@ -0,0 +1,38 @@
...
+
+  const isMalformedUserAgent = !SUPPORTED_OS.includes(deviceOS) || !deviceApplicationVersion;
+  if (isMalformedUserAgent) {
+    return response.failure({ status: 500, message: 'Malformed `User-Agent` header' });
+  }
+
+  const [ssmError, { versions }] = await to(getVersionConfigurations());
+  if (ssmError) {
+    return response.failure(ssmError);
+  }
```
dannil76, line 32: Missing loggin handler

services/version-api/lambdas/getVersionStatus.js
```diff
@@ -0,0 +1,38 @@
...
+ * @param {string} event.headers.User_Agent
+ * @returns {Promise<string>} The version status (i.e "OK", "UPDATE_OPTIONAL" or "UPDATE_REQUIRED")
+ */
+export async function main(event) {
+  const [, deviceApplicationVersion, deviceOS] = event.headers[USER_AGENT].split(SPLIT_SYMBOL);
+
+  const isMalformedUserAgent = !SUPPORTED_OS.includes(deviceOS) || !deviceApplicationVersion;
+  if (isMalformedUserAgent) {
+    return response.failure({ status: 500, message: 'Malformed `User-Agent` header' });
+  }
```
dannil76, line 27: Please add logging handling

services/version-api/serverless.yml
```diff
@@ -0,0 +1,57 @@
...
+      statements:
+        - Effect: Allow
+          Action:
+            - ssm:GetParameter
+          Resource:
+            - !Sub arn:aws:ssm:${self:provider.region}:${AWS::AccountId}:parameter/versionEnvs/${self:provider.stage}
+
+functions:
+  getVersionStatus:
+    handler: lambdas/getVersionStatus.main
```
dannil76, line 47: Maybe the name is somewhat reduntant. The 'Version' is allready in the service name ie. version-api?
I think `getStatus` is enought.

dannil76 (changes requested):

services/version-api/helpers/getVersionConfigurations.js
karlknowit: Removed entirely

karlknowit: re-added. The function throwError fits very well in this case since it throws a generic error message with a status code of 500. The lambda function itself should not throw but instead it should return a http response

dannil76 (approved):

dannil76 (approved):

sepehrnoor (approved):
Looks good to me!

services/version-api/helpers/getApplicationUpdateStatus.js
```diff
@@ -0,0 +1,34 @@
+import smCompare from 'semver-compare';
```
MasterKenth, line 1: `smCompare` seems like an odd abbreviation. `compareSemver`? `compareVersion`?

services/version-api/helpers/getApplicationUpdateStatus.js
```diff
@@ -0,0 +1,34 @@
...
+
+const COMPARISON_RESULT = {
+  EQUAL: 0,
+  SMALLER: -1,
+  GREATER: 1,
+};
+
+/**
+ * @param {{current: string, min: string, max: string}} parameters (properties in semver format i.e "1.2.0")
+ * @returns {string} Status
```
MasterKenth, line 13: `Status` doesn't really say that much. Could we expand this to list the possibilities, or better scope it from `string` to `VERSION_STATUS` constant? (not sure if that is possible with jsdoc, probably not).

services/version-api/helpers/getApplicationUpdateStatus.js
```diff
@@ -0,0 +1,34 @@
...
+ */
+function getApplicationUpdateStatus({ current, min, max }) {
+  const maxComparison = smCompare(current, max);
+  const minComparison = smCompare(current, min);
```
MasterKenth, line 17: Could we rename these to something like `currentComparedToMax` and `currentComparedToMin` respectively? My intention is to make it more clear as when reading the following `if` statement i'm not sure what result to expect.

```javascript
if(currentComparedToMax === COMPARISION_RESULT.SMALLER)
```

Reads more clearly to me as "if current version is smaller than max".

services/version-api/helpers/getVersionConfigurations.js
```diff
@@ -0,0 +1,19 @@
...
+import params from '../../../libs/params';
+import config from '../../../config';
+
+/**
+ * @returns {Promise<{versions: {ios: {min: string, max: string}, android: {min: string, max: string}}}>} SSM parameters
+ */
+async function getVersionConfigurations() {
+  const [error, response] = await to(params.read(config.version.envsKeyName));
+  if (error) {
+    throwError(500);
```
MasterKenth, line 13: `throwError` only throws implicitely. Would it be ok to add a `return` statement afterwards?

```javascript
throwError(500);
return null;
```

Otherwise we can never be 100% sure.

MasterKenth (changes requested):

services/version-api/helpers/getApplicationUpdateStatus.js
karlknowit: I remember I had difficulties naming this constants, but I like the idea about naming them to `currentComparedToMax ` and `currentComparedToMin `. Will do that change

services/version-api/helpers/getApplicationUpdateStatus.js
karlknowit: Will change it to`semverCompare`

services/version-api/helpers/getApplicationUpdateStatus.js
karlknowit: You cannot have a javascript object as type, so will manually typing the return strings. This will also be changed later on when we start using typescript.

PR #316, opened 2022-01-24
fix: navet-ms needs write access to cert bucket

petter-a:
When encrypting the s3 bucket, navet-ms was not updated
with the correct permissions


dannil76 (approved):
Look right to me

petter-a:
Don think this is needed. The initial error was because of wrong SSM Parameters

PR #317, opened 2022-01-24
fix: response format from the version-api now also includes updateUrl

karlknowit:
Changed the return format from getStatus lambda to now also includes updateUrl.

Also moved constants.js file to helpers folder to align with other services.

We do not want the application itself to hold these kind of values in case they would change. By having them specified in backend we could easily change and update them to all environments.

## Explain your solution
the response is now an object containing `status` and `updateUrl` properties instead of only string.

## How to test
1. Checkout this branch
2. run `sls deploy` within service folder for `service x`
3. Make sure that your SSM has versionEnvs configure
4. Run the lambda in postman or application

## Tested environments


helsingborgsstad:
Task linked: [CU-1vkvdnv Ändra response from getStatus lambda i version-api](https://app.clickup.com/t/1vkvdnv)

services/version-api/lambdas/getStatus.js
```diff
@@ -39,8 +39,8 @@ export async function main(event, context) {
     return response.failure(configurationError);
   }
 
-  const { min, max } = configuration.versions[deviceOS];
-  const versionStatus = getApplicationUpdateStatus({ current: deviceApplicationVersion, min, max });
+  const { min, max, updateUrl } = configuration.versions[deviceOS];
+  const status = getApplicationUpdateStatus({ current: deviceApplicationVersion, min, max });
 
-  return response.success(200, { versionStatus });
+  return response.success(200, { status, updateUrl });
```
dannil76, line 45: We should implement 'open api' standards. :)

See how other services implement this. Like in cases-api:

`return response.success(200, {
    type: 'getStatus'
    attributes: {
      status,
      updateUrl,
    },
  });`

dannil76 (changes requested):

services/version-api/lambdas/getStatus.js
karlknowit: Will fix!

dannil76 (approved):
Exelent!

PR #319, opened 2022-01-25
fix: 🐛 issue with expiration time

dannil76:
If case is required for completion and the user starts filling in the form. Case status will be set to 'active:ongoing'.

It casued the lambda to trigger and update expiration time based on the wrong status time frame.

This commit fixes it by not trigger on the state VIVA_COMPLETION_REQUIRED or VIVA_RANDOM_CHECK_REQUIRED

### DIFF
--- a/services/viva-ms/serverless.yml
+++ b/services/viva-ms/serverless.yml
@@ -244,6 +244,8 @@ functions:
           pattern:
             source:
               - casesApi.updateCase
+            detail:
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-1v2tfkw Bugg: Pågående stickprovskontroll raderas efter 72 timmar?](https://app.clickup.com/t/1v2tfkw)

karlknowit (approved):

karlknowit (approved):

MasterKenth (approved):

PR #320, opened 2022-01-28
feat: lambda for monitoring case-states

petter-a:
An internal lambda function that monitors the health of a case in the cases database.

A job is scheduled to run every other hour executing the following flow:
1. Scan the cases table for all cases.
2. For every case, run the tests defined in the validatorList array of src/validators/index.ts
3. If a test fails, write the case number and the error description to a LogGroup in cloudwatch.
4. Finish

The data collected in the cloud watch LogGroup can then be visualised in any way needed. The simplest would be to simply scan the log file periodically or publish the measurements on a dashboard.

The output of a failing test is logged using the following format:
```
{
    testId: 'TEST_HTML_EXPIRED',
    level: 'error',
    message: 'State CASE_HTML_GENERATED not expected after 2 hours.',
  }
```
_The test provided in this PR is fictive. Not sure if it makes sense but it is created as guideline to derive from._

### Add a new test case:
Copy the file src/validators/testTemplate.ts to a new name.
Rename the function and add the logic for checking the error state of the case.
Add the name of the new test to the validatorList array of the file src/validators/index.ts to make the test active.

External dependencies needed to run the test (As with the initial htmlStateExpired test), are provided through the context parameter. Currently the only dependency forwarded to the tests are the ability to stub the system clock using the getAge function:

```
export interface ValidatorContext {
  getAge: (compare: number) => number;
}
```
This interface can be extended as needed.


helsingborgsstad:
Task linked: [CU-1w54we3 Skapa ramverk för testning av ansökan](https://app.clickup.com/t/1w54we3)

dannil76:
Is this branch obsoleted related to what we talked about at the sprint end session?

PR #324, opened 2022-02-04
Fix/jest testing configs

karlknowit:
Refactor to only have one babel config in the repository. Also added a recommendation to install jest-runner plugin for vscode to be able to debug jest tests.

Multiple babel-config files (one in each service) only added more complexity when running tests and working with extensions. By only running the test from root with its configurations makes it easier.

## Explain your solution
Remove the babel.config.js file from each service and point the test script in package.json to root in the repo with the service name as parameter.

## How to test

1. Checkout this branch
2. Run `yarn test` in root, all tests should run
3. or go to a specific service and run `yarn test`, only that service tests will run
4. If you installed the plugin, you can set a breakpoint in the lambda or helper function you want to test. Here is the documentation for jest runner: https://marketplace.visualstudio.com/items?itemName=firsttris.vscode-jest-runner

## Tested environments
Local machine


jlarsson (approved):
Looks good

petter-a (approved):
Looks good!

PR #333, opened 2022-02-14
feat: convert token auth to typescript

petter-a:
Third service to adapt typescript


services/auth/token/src/lambdas/generateToken.ts
```diff
@@ -1,18 +1,40 @@
...
+
+interface JWTToken {
+  personalNumber: string;
+}
+
+interface SDKError {
+  code: string;
+  message: string;
+}
+export const main = async (event: { body: string }, context: { awsRequestId: string }) => {
```
dannil76, line 34: A decision was made to declare functions as normal functions. We use arrow functions for callbacks, for example. map, filter.

We considered it more clear and legible in the code what was a function in this way.

Just like here, there are more places to fix this so it becomes consistent through the code base. It can therefore be good to correct this when we have the code in front of us. :)

dannil76 (changes requested):

services/auth/token/src/lambdas/generateToken.ts
petter-a: fixed

PR #335, opened 2022-02-16
feat: convert bookables-api to typescript

petter-a:
Fourth service to be migrated to typescript


services/bookables-api/src/helpers/searchAdministrators.ts
```diff
@@ -1,16 +1,24 @@
...
+
+async function searchAdministrators(body: {
+  email?: string;
+  mailbox?: string;
+}): Promise<SearchAdministratorsResult | undefined> {
   const [readError, { datatorgetEndpoint, apiKey }] = await to(
     params.read(config.datatorget.envsKeyName)
   );
   if (readError) {
+    // eslint-disable-next-line no-console
```
karlknowit, line 21: perhaps we should take the opportunity to add the log function instead of console.error here?

karlknowit (changes requested):

services/bookables-api/src/helpers/searchAdministrators.ts
petter-a: Well... console log is used everywhere. I'd rather do that in a refactoring task while introducing a log wrapper to the lambas to avoid passing around the was context too ...

PR #337, opened 2022-02-16
feat: convert cases-ms to typescript

petter-a:
6th API converted to typescript


services/case-ms/src/lambdas/updatePdf.ts
```diff
@@ -1,13 +1,25 @@
+/* eslint-disable no-console */
 import to from 'await-to-js';
 
-import config from '../../../config';
+import config from '../libs/config';
```
dannil76, line 4: Not sure if this is a good way to salve it using symlinks?

Please let's have a talk :)

dannil76 (changes requested):

services/case-ms/src/lambdas/updatePdf.ts
petter-a: typescript expects source code to be located in a single folder tree, usually with the root named src. Crosslinking outside this path comes with a lot of caveats and hence symlinking is preferred.

services/case-ms/src/lambdas/updatePdf.ts
```diff
@@ -47,7 +59,7 @@ export async function main(event, context) {
   return true;
 }
 
-function updateCasePdfAttributes(caseKeys, pdf) {
+function updateCasePdfAttributes(caseKeys: CaseKeys, pdf: string) {
```
dannil76, line 62: Option:
If the args have a clear type hinting, could the args itself be trimmed to just keys? :)

dannil76 (changes requested):

services/case-ms/src/lambdas/updatePdf.ts
petter-a: Done

dannil76 (approved):
:)

PR #342, opened 2022-02-22
fix: bankid sign and cancel does not work

petter-a:
attributes not provided to app anymore after refactoring.

Previously, a property named attributes was provided in the response. This property is not coming anymore.


MasterKenth (approved):

dannil76 (approved):
Good

PR #343, opened 2022-02-22
fix: add missing reference to uuid library in package.json

petter-a:
uuid is referenced from code but only installed implicitly. It should
have an explicit include

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,15 +28,15 @@
   },
   "devDependencies": {
     "@babel/core": "^7.16.5",
+    "@babel/eslint-parser": "^7.17.0",
     "@babel/preset-env": "^7.16.11",
     "@babel/preset-typescript": "^7.16.7",
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -28,15 +28,15 @@
...
     "@commitlint/cli": "^11.0.0",
     "@commitlint/config-conventional": "^11.0.0",
-    "@types/node": "^14.14.27",
     "@types/jest": "^27.4.0",
-    "@typescript-eslint/parser": "^5.10.2",
+    "@types/node": "^14.14.27",
     "@typescript-eslint/eslint-plugin": "^5.10.2",
-    "@babel/eslint-parser": "^7.17.0",
+    "@typescript-eslint/parser": "^5.10.2",
     "babel-jest": "^27.5.0",
```
dannil76, line 40: Is this relevant for this PR? :)

petter-a: @dannil76 when issuing yarn add, yarn is sometimes reordering items. The removed and added rows are the same

MasterKenth (approved):

dannil76 (approved):

PR #356, opened 2022-03-08
fix: revert tsconfig setting

petter-a:
The setting module=es6 did not work with local debugging as es6 requires file-extensions
in import statements.


karlknowit (approved):
This is tested by me and Petter before.

PR #366, opened 2022-03-17
feat: allow winston to log to vscode console

petter-a:
While debugging locally, log outputs from winston are suppressed.
This setting will make the output available in the debug console


dannil76 (approved):
Trust you... :)

PR #368, opened 2022-03-17
feat: add extended auto logging

petter-a:
When creating a lambda, wrap the handler with the new log.wrap function to enable start and end logging with extended details automagically.

An example would be
`export const main = log.wrap(async (event, context) => {...`

instead of:
`export const main = async (event, context) => {...`

That is ALL that needs to be done.
The measure will add the following info message types to each request:
"message": "Lambda initialize"
"message": "Lambda finalize"

Each with their own set of metrics.
The initialiser will automatically log the source and detail type of any EventBridge event triggering the lambda.

A complete log-event for a lambda function triggered by EventBridge:
```
{
    "customData": {
        "detailType": "applicationReceivedSuccess",
        "source": "vivaMs.submitApplication"
    },
    "level": "info",
    "message": "Lambda initialize",
    "requestId": "56345ca8-da7c-479f-8e41-1ea8c1bdeaae"
}
```
A complete log-event for a lambda function triggered by API-gateway:
```
{
    "customData": {
        "method": "POST",
        "path": "/auth/bankid/auth",
        "pathParameters": null,
        "queryStringParameters": null,
        "userAgent": "MittHelsingborg/1.1.4/ios/15.2"
    },
    "level": "info",
    "message": "Lambda initialize",
    "requestId": "19303d83-3793-48a6-a6d2-ab71cba56345"
}
```

### How to test
1. Wrap a function with the log.wrap statement as described above
2. Deploy with SLS
3. Invoke the function
4. Verify the CloudWatch log of the lambda to see that the two log entries "Lambda initialise" and "Lambda finalise" is printed

### DIFF
--- a/libs/logs.d.ts
+++ b/libs/logs.d.ts
@@ -1,13 +1,52 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 export default log;
+type Primitive = string | number;
+type LambdaHandler = (event: any, context: any, callback?: any) => Promise<unknown>;
+type Level = 'info' | 'error' | 'warn' | 'verbose' | 'debug';
 
-type LambdaHandler = (event: any, context: any, callback?: any) => Promise<any>;
 declare namespace log {
-  function log(level: any, message: any, requestId: any, errorCode: any, customData?: any): void;
-  function error(message: any, requestId: any, errorCode: any, customData?: any): void;
-  function warn(message: any, requestId: any, errorCode: any, customData?: any): void;
-  function info(message: any, requestId: any, errorCode: any, customData?: any): void;
-  function verbose(message: any, requestId: any, errorCode: any, customData?: any): void;
-  function debug(message: any, requestId: any, errorCode: any, customData?: any): void;
+  function writeLog(level: Level, message: string, customData?: unknown): void;
+  function writeError(message: string, customData?: unknown): void;
+  function writeWarn(message: string, customData?: unknown): void;
+  function writeInfo(message: string, customData?: unknown): void;
+  function writeVerbose(message: string, customData?: unknown): void;
+  function writeDebug(message: string, customData?: unknown): void;
+  function log(
+    level: Level,
+    message: string,
+    requestId: string,
+    errorCode: Primitive,
+    customData?: unknown
+  ): void;
+  function error(
+    message: string,
+    requestId: string,
+    errorCode: Primitive,
+    customData?: unknown
+  ): void;
+  function warn(
+    message: string,
+    requestId: string,
+    errorCode: Primitive,
+    customData?: unknown
+  ): void;
+  function info(
+    message: string,
+    requestId: string,
+    errorCode: Primitive,
+    customData?: unknown
+  ): void;
+  function verbose(
+    message: string,
+    requestId: string,
+    errorCode: Primitive,
+    customData?: unknown
+  ): void;
+  function debug(
+    message: string,
+    requestId: string,
+    errorCode: Primitive,
+    customData?: unknown
+  ): void;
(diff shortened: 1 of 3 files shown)


libs/logs.d.ts
```diff
@@ -1,10 +1,64 @@
...
-export default log;
+import {
+  APIGatewayProxyEvent,
+  Context,
+  APIGatewayProxyResult,
+  APIGatewayProxyCallback,
+} from 'aws-lambda';
+
+type LambdaHandler = (
+  event: APIGatewayProxyEvent,
```
MasterKenth, line 9: `event` isn't always `APIGatewayProxyEvent` (same for other variables). Probably an auto-generation error?

libs/logs.js
```diff
@@ -16,25 +25,93 @@ const log = {
...
+    return (event, context, callback) => {
+      log.initialize(event, context);
+
+      const executor = lambda(event, context, (error, response) => {
+        log.finalize(response, error);
+        callback(error, response);
+      });
+
+      if (!(executor instanceof Promise)) {
+        return undefined;
```
MasterKenth, line 98: According to the typings `wrap` should always return (a function that returns) a `Promise`, never `undefined`. Could we return an empty Promise (`Promise.resolve()`), or maybe bake this check into the `Promise` body instead.

libs/logs.js
```diff
@@ -16,25 +25,93 @@ const log = {
...
+
+  finalize: (response, error) => {
+    log.writeInfo('Lambda finalize', {
+      statusCode: response?.statusCode ?? 0,
+      error,
+    });
+  },
+
+  wrap: lambda => {
+    return (event, context, callback) => {
```
MasterKenth, line 89: Since we're dealing with promises anyway I think it would be cleaner to rewrite `wrap` to `return async (event, context, callback) => {...}` and utilize `await` instead.

libs/logs.js
```diff
@@ -7,7 +7,16 @@ const logger = winston.createLogger({
   exitOnError: false,
 });
 
+let requestId = 'N/A';
+
 const log = {
+  writeLog: (level, message, customData = {}) => {
```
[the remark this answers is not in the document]
petter-a: The reason being that the existing functions requires you to add an errorCode (Which is redundant) and also a "requestId" that could instead will be provided automatically through the wrapper

libs/logs.js
```diff
@@ -7,7 +7,16 @@ const logger = winston.createLogger({
   exitOnError: false,
 });
 
+let requestId = 'N/A';
```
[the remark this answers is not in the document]
petter-a: Fixed

libs/logs.d.ts
petter-a: Fixed

libs/logs.js
petter-a: Fixed change to empty return statement (void) .... Following the generic Handler typings the return statements can be:
void | Promise<TResult>

libs/logs.js
petter-a: Not sure the definition of a Handler should be:
type Handler<TEvent = any, TResult = any> = (event: TEvent, context: Context, callback: Callback<TResult>) => void | Promise<TResult>

libs/logs.js
```diff
@@ -16,25 +25,93 @@ const log = {
...
+    });
+  },
+
+  wrap: lambda => {
+    return (event, context, callback) => {
+      log.initialize(event, context);
+
+      const executor = lambda(event, context, (error, response) => {
+        log.finalize(response, error);
+        callback(error, response);
```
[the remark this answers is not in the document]
petter-a: Fixed

kajm (approved):

PR #372, opened 2022-03-25
Feat/refactor auth token

petter-a:
Refactor and adding logging + unittests


services/auth/token/src/lambdas/authorize.ts
```diff
@@ -1,42 +1,35 @@
...
+}
+/* istanbul ignore next */
+export const main = log.wrap(async event => {
+  return await lambda(event, {
+    getSecret: secrets.get,
+    verifyToken,
+  });
+});
+
+export async function lambda(event: LambdaEvent, lambdaContext: LambdaContext) {
```
kajm, line 25: Shouldn't we use try/catch in the lambda function?

services/auth/token/src/lambdas/generateToken.ts
```diff
@@ -16,194 +15,125 @@ interface SecretsConfig {
...
-    validateToken(grantTypeValues?.secretsConfig, grantTypeValues?.token ?? '')
-  );
-
-  if (validateTokenError) {
-    log.warn(
-      'Validate token error',
-      context.awsRequestId,
-      'service-auth-token-generateToken-003',
-      validateTokenError
+    const accessToken = await generateToken(
```
kajm, line 54: **OPTIONAL**

Since the creation of `refreshToken` is not dependent on `accessToken`, we could just put those requests in a `Promise.all` like:

```js
    const [accessToken, refreshToken] = await Promise.all([
      generateToken(
        CONFIG_AUTH_SECRETS.accessToken,
        decodedGrantToken.personalNumber,
        ACCESS_TOKEN_EXPIRES_IN_MINUTES,
        lambdaContext
      ),
      generateToken(
        CONFIG_AUTH_SECRETS.refreshToken,
        decodedGrantToken.personalNumber,
        REFRESH_TOKEN_EXPIRES_IN_MINUTES,
        lambdaContext
      )
    ]);
```

kajm (changes requested):
Feels like this PR should be rebased since there are a lot of files shown that shouldn't be, like booking and upload pdf etc.

services/auth/token/src/lambdas/generateToken.ts
kajm: ... and even:

```js

const generateTokenBySecretConfig = (config) => generateToken(
  config, 
  decodedGrantToken.personalNumber, 
  ACCESS_TOKEN_EXPIRES_IN_MINUTES, 
  lambdaContext 
);

const [accessToken, refreshToken] = await Promise.all([ 
  generateTokenBySecretConfig(CONFIG_AUTH_SECRETS.accessToken), 
  generateTokenBySecretConfig(CONFIG_AUTH_SECRETS.refreshToken), 
]);
```

dannil76:
> Feels like this PR should be rebased since there are a lot of files shown that shouldn't be, like booking and upload pdf etc.

Done!

PR #380, opened 2022-04-12
Feat/handle new application on submit

kajm:
- Handle new application in submitApplication lambda
- Added tests for submitApplication lambda
- Refactored to the new agreed format
- Moved out functions used in the lambda file to other helper files

Before, the submitApplication lambda could only handle recurring cases, but there is a need to also handle `new` cases.

## Explain your solution
The lambda now also checks if the `currentFormId` in a case is for a recurring form or a new application form. Depending on the current form used by the case, parameters sent to viva is different.

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for `service viva-ms`

## Tested environments


helsingborgsstad:
Task linked: [CU-258zqvv submittApplication-lambda ska kunna skicka case för grundansökan till Vada](https://app.clickup.com/t/258zqvv)

PR #395, opened 2022-05-03
feat: add lib to jest coverage ignore 

petter-a:
Coverage is now ignoring the /libs/ directory making lambda testing easier

coveragePathIgnorePatterns: ['/libs/'],

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,6 +7,7 @@ module.exports = {
       babelConfig: 'babel.config.js',
     },
   },
+  coveragePathIgnorePatterns: ['/libs/'],
   coverageThreshold: {
     global: {
       branches: 80,


services/bookables-api/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "test": "cd ../../ && node node_modules/.bin/jest bookables-api"
+    "test": "cd ../../ && node node_modules/.bin/jest bookables-api --coverage"
```
kajm, line 7: Do we really need the --coverage option? The developer could add it by himself when running the script like:

`yarn test --coverage`

petter-a: fixed

kajm (approved):

PR #405, opened 2022-05-24
feat: add form ids for new application

kajm:
CreateNewVivaCase lambda were missing form ids for random check and completions for new application, hence adding those.

When a new case for new application is created, the default value for answers, encryption and position is missing for random check and completions. Therefore they needed to be added.

## Explain your solution
Added the missing `newApplicationRandomCheckFormId` and `newApplicationCompletionFormId` in the formIdList which creates the default values in DynamoDb.

## How to test

Example when running the lambda locally:
1. Make sure that your test person does not have any cases in the cases table
2. Make sure that you have the `newApplicationCompletionFormId` and `newApplicationRandomCheckFormId` ssm parameters in vivaEnvs in AWS SSM
3. Run the lambda function
4. Check the created case in dynamoDb table and make sure you have 3 forms with default values in it

### DIFF
--- a/services/viva-ms/src/lambdas/createNewVivaCase.ts
+++ b/services/viva-ms/src/lambdas/createNewVivaCase.ts
@@ -60,14 +60,19 @@ export async function lambda(
   lambdaContext: LambdaContext
 ): Promise<boolean> {
   const { user } = event.detail;
-  const { newApplicationFormId } = await params.read(config.cases.providers.viva.envsKeyName);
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-25zu6hr Vid skapandet av grundsansökan ska completion-form-id läggas till](https://app.clickup.com/t/25zu6hr)

dannil76 (approved):

PR #409, opened 2022-05-25
fix: 🐛 (viva-ms) broken tests

dannil76:

The tests is now up to date with the fix regarding the new way of
checking if case isLocked

## How to test

Concrete example:

1. Checkout this branch
2. run `yarn test` within service folder for `service viva-ms`

## Tested environments

- [x] Your local environment.

### DIFF
--- a/services/viva-ms/test/helpers/caseDecision.test.js
+++ b/services/viva-ms/test/helpers/caseDecision.test.js
@@ -23,7 +23,7 @@ const workflow = {
     completiondescription: null,
     completionduedate: null,
     islockedwithoutcompletionreceived: null,
-    islocked: '2022-02-02T15:42:22+01:00',
+    islocked: null,
   },
 };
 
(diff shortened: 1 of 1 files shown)


services/viva-ms/test/helpers/caseDecision.test.js
```diff
@@ -154,7 +154,29 @@ it('Before the Viva administrator processes the application the case status must
 });
 
 it(`Calculation exists in the Viva workflow collection, then status type must be ${ACTIVE_PROCESSING}`, () => {
-  const results = decideNewCaseStatus(workflowCalculation);
+  const workflow = {
```
kajm, line 157: You have already declared `workflow` constant in a more global scope. Consider renaming this one to avoid conflicts.

And also, a little cleanup

```js
const yourConstantName = {
   ...workflow,
   islocked: '123',
  calculations: {
      calculation: {},
    },
}
```

kajm (changes requested):

kajm (approved):

PR #423, opened 2022-07-19
Fix/handle undefined address

kajm:
Handle navet users which does not have an address registered in Navet

If a user didn't had an address registered in Navet and tried to login to the application, the user would be signed out immediately. That wasn't the wanted behavior, so instead on throwing in the `pollUser` lambda, we now handle addresses which doesn't exists.

## Explain your solution
The lambda now defaults the addresses properties to undefined if they do not exists

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for `service navet-ms`
3. Make sure you have a user registered in Navet that you can test with
4. Remove user if exists in the DynamoDB table
5. Login to the application, the user should appear in the DynamoDB table
6. Check the profile page in the application and make sure everything except address are filled

## Tested environments


helsingborgsstad:
Task linked: [CU-23fqf1y Bugg: Som invånare ska jag kunna komma vidare i appen även ifall det inte finns en adress registrerad i Navet](https://app.clickup.com/t/23fqf1y)

services/navet-ms/src/helpers/parser.ts
```diff
@@ -46,14 +56,10 @@ export const getPersonPostCollection = xml =>
...
-    try {
-      const parsedOnce = xml.split('<faultstring>');
-      const parsedTwice = parsedOnce[1].split('</faultstring>');
-      resolve(parsedTwice[0]);
-    } catch (error) {
-      reject(error);
-    }
-  });
+export function getErrorMessageFromXML(xml: string): string {
+  const parsedOnce = xml.split('<faultstring>');
```
MasterKenth, line 62: add ? pls

MasterKenth (changes requested):

MasterKenth (approved):

PR #429, opened 2022-08-23
Fix/mixed

dannil76:

Cleanups


libs/certificates.js
```diff
@@ -1,4 +1,4 @@
-import AWS from './aws-sdk';
+import AWS from 'aws-sdk';
```
kajm, line 1:
```js
import S3 from 'aws-sdk/clients/s3';

const s3Client = new S3({apiVersion: '2006-03-01'});
```

libs/dynamoDb.js
```diff
@@ -1,4 +1,4 @@
-import AWS from './aws-sdk';
+import AWS from 'aws-sdk';
```
kajm, line 1:
```js
import DynamoDB from "aws-sdk/clients/dynamodb";

const dynamoDBClient = new DynamoDB.DocumentClient();
```

kajm (approved):

PR #445, opened 2022-09-16
Refactor/viva ms helper ts

dannil76:
Minor TS related cleanups

Make linter happy and follow BP

## How to test
Nothing


MasterKenth (approved):

services/viva/helpers/populateForm.ts
```diff
@@ -0,0 +1,61 @@
...
+
+        const [childAnswer] = childAnswerList;
+        return childAnswer;
+      })
+      .filter(Boolean);
+
+    return [...answers, ...childAnswers];
+  }, [] as CaseFormAnswer[]);
+
+  return answers;
```
kajm, line 45: This block of code is very hard to read. One solution would be to split every callback function used in .map to a separate function

kajm (approved):

services/viva/helpers/populateForm.ts
dannil76: will fix in later PR

PR #446, opened 2022-09-16
Refactor/users api typescript rewrite

kajm:
Rewrite lambda functions to typescript and added tests

We want our lambdas to be written in typescript to improve development. Also added tests for all lambdas.

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for `service users-api


services/users-api/src/helpers/types.ts
```diff
@@ -0,0 +1,17 @@
+export interface CaseUser {
```
kajm, line 1: Should we really have CaseUser here? This is a duplicate ...

kajm: @dannil76 @MasterKenth

dannil76 (approved):

MasterKenth (approved):

PR #449, opened 2022-09-26
Feat/encryptionKeyId

MasterKenth:

Added a new variable `encryptionKeyId` that works the same as the current `symmetricKeyName`.

`symmetricKeyName` has become a misleading name as it no longer has anything to do with symmetric key. Just renaming it will break existing cases however, so we introduce a new variable that will take its place once it has been fully adopted (should only take around 1 case period).

## How to test

Concrete example:

1. Checkout this branch
2. Generate a new case
3. Verify it generates a `encryptionKeyId` variable next to the `symmetricKeyName`, with the same value

### DIFF
--- a/services/cases-api/src/helpers/schema.js
+++ b/services/cases-api/src/helpers/schema.js
@@ -23,6 +23,7 @@ const encryptedAnswers = Joi.object({
 
 const encryption = Joi.object({
   type: Joi.string().required(),
+  encryptionKeyId: Joi.string(),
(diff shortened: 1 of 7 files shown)


helsingborgsstad:
Task linked: [CU-3gpdqvd Lägg till encryptionKeyId och använd istället för symmetricKeyName](https://app.clickup.com/t/3gpdqvd)

kajm (approved):

dannil76 (approved):

PR #453, opened 2022-10-05
Fix/event flow controll

dannil76:
Fixed so that completions status is set correct when application is of type 'new'

### DIFF
--- a/services/viva/helpers/completions.ts
+++ b/services/viva/helpers/completions.ts
@@ -9,6 +9,7 @@ import {
   ACTIVE_COMPLETION_REQUIRED_VIVA,
   ACTIVE_COMPLETION_SUBMITTED_VIVA,
   ACTIVE_SUBMITTED,
+  ACTIVE_NEW_APPLICATION_RANDOM_CHECK_VIVA,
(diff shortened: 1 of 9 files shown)


helsingborgsstad:
Task linked: [CU-3jnb5fg Bugg - ärendet i viva synkas inte med ärendet i dynamoDB](https://app.clickup.com/t/3jnb5fg)

services/viva/helpers/completions.ts
```diff
@@ -18,54 +18,95 @@ import {
...
-  }
+  const defaultResult: CompletionsResult = {
+    statusType: ACTIVE_COMPLETION_REQUIRED_VIVA,
+    state: VIVA_COMPLETION_REQUIRED,
+  };
 
-  if (isRandomCheck(completions)) {
-    return VIVA_RANDOM_CHECK_REQUIRED;
-  }
+  const result = rules.reduce<CompletionsResult | undefined>(
```
MasterKenth, line 97: Could replace reducer with `const result = rules.find(completionRule => completionRule.condition(params));`.

PR #109, opened 2020-11-12
Feat/viva ms x api key

dannil76:

PR #459, opened 2022-10-24
feat(createnewvivacase): only allowed users to create new viva case

kajm:
Only allow approved new application users to create a new viva case

There is a need to only test new application functionality on specific users, hence adding this check.

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for `service viva`
3. Add a personal number in SSM (https://github.com/helsingborg-stad/helsingborg-io-sls-resources/pull/109)


services/viva/microservice/src/lambdas/createNewVivaCase.ts
```diff
@@ -146,5 +157,6 @@ export const main = log.wrap(event => {
     readParams: params.read,
     getUserCasesCount,
     getTemplates: getFormTemplates,
+    getApprovedNewApplicationUsers: getApprovedNewApplicationUsers,
```
dannil76, line 160: can be trimmed to just getApprovedNewApplicationUsers,

dannil76 (changes requested):

dannil76 (approved):

PR #462, opened 2022-10-26
fix: add period start date on new application

dannil76:
Add current application date to period start date

To correctly display month in form

### DIFF
--- a/services/viva/helpers/createCase.ts
+++ b/services/viva/helpers/createCase.ts
@@ -1,19 +1,18 @@
 import uuid from 'uuid';
 
-import {
+import { VivaPersonType } from '../types/vivaMyPages';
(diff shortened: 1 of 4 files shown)


services/viva/helpers/createCase.ts
```diff
@@ -114,6 +114,11 @@ function getInitialFormAttributes(
   );
 }
 
+function createPeriodStartDate(): number {
+  const nowDate = new Date();
+  return new Date(nowDate.getFullYear(), nowDate.getMonth(), nowDate.getDate()).getTime();
```
MasterKenth, line 119: Use Date.UTC instead.

MasterKenth (changes requested):
Tests failing - seems to be timezone issues.

PR #469, opened 2022-11-07
feat(lambdawrapper): decode authorization token if exists

kajm:
The authorization token were missing in the input transformer and in the event to the lambda function, causing the decoding to fail in the function. Now we decode the token directly in the input transformer and put the result (personal number) directly in the input event to the lambda.

Some lambda functions are dependent on a users personal number in order to request the right data. If the personal number isn't included in the function input, then the requests will fail.

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for `service users-api` as an example
3. Run a lambda function that includes the wrapper for restJSON and check the input to the lambda
4. ... or run the lambdawrapper tests

### DIFF
--- a/libs/lambdaWrapper.test.ts
+++ b/libs/lambdaWrapper.test.ts
@@ -1,11 +1,13 @@
 import type { APIGatewayEvent, Context } from 'aws-lambda';
 import { wrappers } from './lambdaWrapper';
 import log from './logs';
+import jwt from 'jsonwebtoken';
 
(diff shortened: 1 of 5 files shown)


helsingborgsstad:
Task linked: [CU-3rcgpbk Decod:a token i inputtransformern](https://app.clickup.com/t/3rcgpbk)

MasterKenth (approved):

PR #477, opened 2022-11-21
fix(attachment.ts): use correct property for viva attachment name

kajm:
The api were sending the wrong name for attachments to VIVA, making VIVA display the uploaded id instead of the external display name.

The id of an attachment should not be (and is not a good name) displayed to administrators in VIVA. The uploaded id is only for EKB API.

## How to test

1. Checkout this branch
2. run `sls deploy` within service folder for `service viva microservice`
3. Sen in an application to VIVA with attachments


helsingborgsstad:
Task linked: [CU-3vtacyb Bugg: attachments skickas in med fel namn till viva](https://app.clickup.com/t/3vtacyb)

PR #487, opened 2022-12-09
fix(html): use correct template function and hbs file for new applications

MasterKenth:

Fixed function checked for key when it should check for value.

Fix use correct template for new applications.

## How to test

Concrete example:

1. Checkout this branch
2. Run through new application
3. Verify it generates correct PDF


dannil76 (approved):

dannil76:
In med det!

PR #490, opened 2022-12-14
Fix/allow new application change

kajm:
Refactored `createNewVivaCase` lambda to not create new viva cases when its not supposed to. Implemented logic which follows the decide chart:

image

The grundansökan-button did appear when not supposed to, which meant that a new application case where created. This PR hopefully fixes the issue and avoid creating new application cases when not needed.

## How to test

Concrete example:

1. Checkout this branch
2. run `sls deploy` within service folder for `service viva`
3. Run the EKB application and add a co-applicant before answering the grundansökan form
4. Login as the co-applicant
5. The grundansökan button should not be visible
6. no new cases should either be created in cases table

### DIFF
--- a/services/viva/helpers/newApplication.ts
+++ b/services/viva/helpers/newApplication.ts
@@ -1,0 +1,10 @@
+import type { VivaParametersResponse } from '../types/ssmParameters';
+
+export function isFormNewApplication(
+  formIds: VivaParametersResponse,
+  targetFormIdValue: string
+): boolean {
+  return Object.entries(formIds).some(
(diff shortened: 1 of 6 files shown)


helsingborgsstad:
Task linked: [CU-3xjhe0r Bugg: Knapp för grundansökan visas för medsökande](https://app.clickup.com/t/3xjhe0r)

dannil76 (changes requested):

dannil76:
Något som du @kajm tycker är oklart med mina kommentarer? Tänker att min review legat ett tag? :)

dannil76 (approved):

PR #494, opened 2022-12-21
refactor(openapi): add docs for status messages

kajm:
Added openapi documentation for status api

We want our api:s to be documented, hence creating this openapi docs.

## How to test

1. Checkout this branch
2. Open status-api with VsCode
3. runt the openapi.yml file in swagger

### DIFF
--- a/services/status-api/openapi.yml
+++ b/services/status-api/openapi.yml
@@ -21,14 +21,15 @@ paths:
         - Status
       summary: Get API status messages.
       description: |
-        This endpoint returns a list of messages that would be used and displayed by a consumer.
-        The list of messages returned in the response could be on of the following:
+        This endpoint returns a filtered list of message objects and includes only messages that are currently valid
+        based on their start and expiry dates.
 
-        1. info
-        2. warning
-        3. maintenance
+         The list of messages returned in the response could be in one of the following 
+         types (starting with the lowest severity first):
 
-        in which the severity of the message is increased with start from position1.
+         1. info
+         2. warning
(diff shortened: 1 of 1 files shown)


helsingborgsstad:
Task linked: [CU-3xcre5g Skapa OpenAPI för status service](https://app.clickup.com/t/3xcre5g)

services/status-api/openapi.yml
```diff
@@ -0,0 +1,103 @@
...
+
+paths:
+  /status/messages:
+    get:
+      tags:
+        - Status
+      summary: Get API status messages.
+      description: |
+        This endpoint returns a list of messages that would be used and displayed by a consumer.
+        The list of messages returned in the response could be on of the following:
```
MasterKenth, line 25: Should refrain from using suggestive actions I think, so skip the "that would be used..." part and simply just explain how this endpoint works (inputs and outputs).

Could also probably rephrase this to something like "in order of severity:".

services/status-api/openapi.yml
```diff
@@ -0,0 +1,103 @@
...
+
+        1. info
+        2. warning
+        3. maintenance
+
+        in which the severity of the message is increased with start from position1.
+
+      security:
+        - ApiKeyAuth: []
+      parameters:
```
MasterKenth, line 35: This endpoint does not require a JWT (only api key).

MasterKenth (changes requested):
comments.

MasterKenth (approved):

PR #515, opened 2023-02-14
fix(createvivacase): use previous case answer instead of applicant info if possible

MasterKenth:

Prioritized using answers from any previous case over using user information when pre-populating address, phone etc.

The last case if any holds the most up-to-date information as the users table info is not updated.

## How to test

Concrete example:

1. Checkout this branch
2. Delete any existing cases and create a new case for an applicant
3. Verify the created case and form have answers for address etc. that matches the info in the users table
4. Fill out the case with a different address and submit
5. Create a new case for the same user
6. Verify the new case form have answers for address etc. that matches the info provided in the last case

### DIFF
--- a/services/viva/microservice/test/lambdas/createVivaCase.test.ts
+++ b/services/viva/microservice/test/lambdas/createVivaCase.test.ts
@@ -26,7 +26,6 @@ import type {
 } from '../../src/types/caseItem';
 
 const MOCK_DATE = dayjs('2022-01-01T00:00:00Z');
-jest.useFakeTimers('modern').setSystemTime(MOCK_DATE.unix() * 1000);
 
 const mockUuid = '00000000-0000-0000-0000-000000000000';
 jest.mock('uuid', () => ({ v4: () => mockUuid }));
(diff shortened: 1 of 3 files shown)


helsingborgsstad:
Task linked: [CU-3n5jkpg Vid ihopslagning av data från Users och formulär, prioritera det som tidigare angivits i formuläret](https://app.clickup.com/t/3n5jkpg)

dannil76 (approved):

services/viva/microservice/test/lambdas/createVivaCase.test.ts
```diff
@@ -174,6 +185,83 @@ function createLambdaInput(persons: VivaPersonsPerson | null = null): LambdaRequ
...
+          [readParametersResponse.completionFormId]: defaultFormProperties,
+        }),
+      getPeriodConfig: getMockPeriodConfig,
+    },
+    value
+  );
+}
+
 describe('createVivaCase', () => {
   beforeAll(() => {
```
dannil76, line 266: Behöver du denna? Finns redan på rad 29.

dannil76 (changes requested):

